PR #1654, opened 2020-02-10
fix(orbitdb): Fix orbitdb stream logic

gfanton:
- subscribe to event instead of listing orbitdb log each second
- improve stream test


moul:
@n0izn0iz @gfanton from what I understand, the plan is to close this PR, can one of you explain why we don't want to use this code?

n0izn0iz:
mainly because in it's current state it adds a corner case where we can loose events silently so in the end it needed too much attention although:
- it's in something that will be replaced by the real protocol
- it's an improvement of a hack that is there because the orbitdb stream function is not behaving as expected so we decided to fix the orbitdb function directly instead "improving" the hack
- the bug in the initial hack that triggered this PR is fixed on master and it's passing the test added in this PR (which was added to master too)

PR #1665, opened 2020-02-18
chore(js): update pbhbs

gponsinet:
depends on #1652


moul:
@gponsinet is this PR still relevant?

gponsinet:
If you have no problem on master you should close it

moul:
cc @n0izn0iz @clegirar what do you think we should do?

n0izn0iz:
Closing since I don't think we have problems on master

PR #1748, opened 2020-03-19
feat(tool): berty mini cluster with tmux

moul:
This PR bootstraps the mini tool `./tool/berty-mini-cluster` which allows starting `berty mini` under various contexts (with/without docker, with/without network bridge, etc)

The current state is relatively poor in term of use cases, but can easily be updated as soon as we find new testing ideas


moul:
This PR keeps generating 500 errors on GitHub, just recreated a new one -> #1792

PR #1756, opened 2020-03-25
feat(chat): use jdenticon as user profile picture

### LINKED ISSUE #1714: Use procedural user pictures
We should discuss what to do when we have no picture, ideas:

- Show a question mark
- Show a procedurally generated image from an id depending on the context (memberPk for a member, contactPk for a contact, etc..)

[EDIT]
After discussion we decided to use procedurally generated pictures

The picture should be based on the account public key for the user and contacts

Use this library: https://arkvatar.com

[EDIT2]

Unfortunately, arkvatar is a service only so we decided to use jdenticon (https://www.npmjs.com/package/react-native-jdenticon)

n0izn0iz:
Fixes #1714


n0izn0iz:
Closed in favor of #1768

PR #1780, opened 2020-03-26
fix(chat): align bullet point

### LINKED ISSUE #1712: Align bullet point lists
Screenshot 2020-03-10 at 21 52 38
Screenshot 2020-03-10 at 20 52 56
Screenshot 2020-03-10 at 21 54 01

clegirar:
Fixes #1712


n0izn0iz:
there is bullet point not only in scan, in other components aswell (some settings at least I think)

n0izn0iz:
I just saw that in the "app updates" page there is an unaligned bullet list, was it fixed in this PR?

PR #1784, opened 2020-03-26
fix: launch error because of bluetooth privilege required

D4ryl00:
When launching the app, an error occurred at startup:
```
This app has crashed because it attempted to access privacy-sensitive data without a usage description.  The app's Info.plist must contain an NSBluetoothAlwaysUsageDescription key with a string value explaining to the user how the app uses this data.
```

According to this error message, I add that key to Info.plist.


aeddi:
For the information of those reading this PR: this fix is required by iOS 13.

PR #1785, opened 2020-03-26
WIP: More v2 paths fixes

n0izn0iz:


.buildkite/pipeline.yml
```diff
@@ -16,7 +16,7 @@ steps:
       - n0izn0iz/docker#v3.5.4:
           image: bertytech/protoc:22
           propagate-uid-gid: true
-          workdir: /go/src/berty.tech/berty
+          workdir: /go/src/berty.tech/berty/v2
```
moul: The way gomodules are working for major version >=2 is not exactly mapping the filesystem, in our case,  you should not have v2 in your path, but only in the "go.mod" file and in the go imports

.buildkite/pipeline.yml
```diff
@@ -134,7 +134,7 @@ steps:
       queue: "macos"
     plugins:
       - gopath-checkout#v1.0.1:
-          import: berty.tech/berty
+          import: berty.tech/berty/v2
```
moul: this one is maybe good, I don't know how this plugin is working

api/bertydemo.proto
```diff
@@ -4,7 +4,7 @@ package berty.protocol;
 
 import "github.com/gogo/protobuf/gogoproto/gogo.proto";
 
-option go_package = "berty.tech/berty/go/pkg/bertydemo";
+option go_package = "berty.tech/berty/v2/go/pkg/bertydemo";
```
moul: nope, you should keep real file system destination

in most of the cases it will work without any hack

the only exception I found was when a .proto file imports another one (and only on current gogoproto compiler version), then you need to apply a `sed` after the `protoc` command (see the current go/Makefile file as an example)

.buildkite/pipeline.yml
n0izn0iz: thanks for the explanations

n0izn0iz:
Closing since what triggered this PR is #1801 and it was due to an old gomobile version, the js changes here are now in #1802

PR #1796, opened 2020-03-27
refactor: cleanup dead errcodes and rename some of them for consistency

aeddi:
This PR:
- Renames some errcodes for consistency
- Reorganizes errcodes in relevent categories
- Removes the unused errcodes


api/errcode.proto
```diff
@@ -9,80 +9,73 @@ option go_package = "berty.tech/berty/go/pkg/errcode";
...
-  ErrBridgeNotRunning = 2002;
+  TODO = 666; // indicates that you plan to create an error later
+  ErrNotImplemented = 777; // indicates that a method is not implemented yet
+  ErrInternal = 888; // indicates an unknown error (without Code), i.e. in gRPC
+
+  //----------------
+  // Generic errors
+  //----------------
+
+  // Parameters and I/O errors
```
moul:
```suggestion
  // Parameters and I/O errors

```

api/errcode.proto
```diff
@@ -9,80 +9,73 @@ option go_package = "berty.tech/berty/go/pkg/errcode";
...
+  ErrInvalidInput = 100;
+  ErrMissingInput = 101;
+  ErrSerialization = 102;
+  ErrDeserialization = 103;
+  ErrStreamRead = 104;
+  ErrStreamWrite = 105;
+  ErrMissingMapKey = 106;
+
+  // Crypto errors
```
moul:
```suggestion
  // Crypto errors

```

api/errcode.proto
```diff
@@ -9,80 +9,73 @@ option go_package = "berty.tech/berty/go/pkg/errcode";
...
+  ErrCryptoRandomGeneration = 200;
+  ErrCryptoKeyGeneration = 201;
+  ErrCryptoNonceGeneration = 202;
+  ErrCryptoSignature = 203;
+  ErrCryptoSignatureVerification = 204;
```
moul:
```suggestion
  ErrCryptoVerification = 204;
```

???

not sure about this one :)

api/errcode.proto
```diff
@@ -9,80 +9,73 @@ option go_package = "berty.tech/berty/go/pkg/errcode";
...
+  ErrCryptoDecrypt = 205;
+  ErrCryptoEncrypt = 206;
+  ErrCryptoKeyConversion = 207;
+
+  //-----------------
+  // Specific errors
+  //-----------------
+
+  // OrbitDB errors
```
moul:
```suggestion
  // OrbitDB errors

```

etc, same for the next ones

api/errcode.proto
aeddi: Not sure either ^^

I also hesitated to put only `Verification` instead of `SignatureVerification` but by checking a little on the web, usually we talk about signature verification to contextualize.

@glouvigny can you give us your opinion?

aeddi: Looks like `Verification` in context of cryptography could mean a lot of things besides signatures: https://www.google.com/search?lr=lang_en&q=cryptography+verification

I'll keep it as is if you don't mind. ;)

PR #1804, opened 2020-03-30
Some Makefile rules doesn't works inside the go folder

gfanton:
Some makefile rules that depends on bazel doesn't works anymore
fix #1804


moul:
@n0izn0iz can you give a look?

n0izn0iz:
This looks ok to me but we loose several basic rules, we could use `make -C ..` instead if we want to keep them

PR #1815, opened 2020-03-31
fix(chat): fix message border styles

### LINKED ISSUE #1737: Comic balloon style for messages
Outward bottom corner should look like this 
Screenshot 2020-03-19 at 19 04 30

clegirar:
Fixes #1737

### DIFF
--- a/js/packages/berty-storybook/chat/shared-components/Message.tsx
+++ b/js/packages/berty-storybook/chat/shared-components/Message.tsx
@@ -94,16 +94,31 @@ export const Message: React.FC<MessageProps> = ({
 				<View
 					style={[
 						padding.small,
-						border.radius.medium,
+						border.radius.top.medium,
+						payload.isMe ? border.radius.left.medium : border.radius.right.medium,
 						styleMsg,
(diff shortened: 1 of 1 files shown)


js/packages/berty-storybook/chat/shared-components/Message.tsx
```diff
@@ -111,10 +112,22 @@ export const Message: React.FC<MessageProps> = ({
...
+							</Text>
+							{payload.isMe && (
+								<Icon
+									name={payload.acknowledged ? 'navigation-2' : 'navigation-2-outline'}
+									width={12}
+									height={12}
+									fill={color.blue}
+								/>
+							)}
+							<Text style={[text.color.grey, padding.left.scale(3), _styles.dateMessage]}>
```
n0izn0iz: I think we should remove the text if we add the icon

clegirar: I think it's good if we have both for more clarity, if we have just an icon the user may have a different understanding from ours no ?

n0izn0iz: my bad I had bad memory of the mockups

PR #1824, opened 2020-04-02
WIP feat: extract contact_pk from the contact request reference

moul:
This allows keeping short QR codes while having the contact PK which is used to generate avatars

This idea comes from a discussion with @n0izn0iz, @clegirar and @zooma2k18 

@berty/go-owners what's your opinion on this?

### DIFF
--- a/api/bertytypes.proto
+++ b/api/bertytypes.proto
@@ -496,11 +496,13 @@ message ContactRequestResetReference {
 
 message ContactRequestSend {
   message Request {
-    // reference is an opaque message describing how to connect to the other account
-    bytes reference = 1;
+    bytes contact_pk = 1 [(gogoproto.customname) = "ContactPK"];
+    
+    // request_reference is an opaque message describing how to connect to the other account
+    bytes request_reference = 2;
 
     // contact_metadata is the metadata specific to the app to identify the contact for the request
-    bytes contact_metadata = 2;
+    bytes contact_metadata = 3;
   }
   message Reply {}
 }


glouvigny:
`ShareableContact` already exists (actually its the deserialized version of reference) and will do what you want I guess, but in that case it will be up to the client to serialize this message to render URLs and QR Codes.

moul:
@n0izn0iz can you check if `ShareableContent` is what you need and if I can close this PR?

n0izn0iz:
I would really like to have a call that takes the opaque reference as input and returns the ShareableContact
This would allow us to make this call when we scan a QR or open a deep link and rely on bertyprotocol do the format checking

In the UX designs currently when we scan a QR we show this screen before actually sending the request (calling contactRequestSend):

Screenshot 2020-04-04 at 00 41 45


Without a way to check the opaque reference in the protocol we would need to know the format of the public keys and the rdv reference in the front to check it and print an error if malformed before actually trying to send the request (for example too much bytes or not enough)

Another way to solve this in front only would be to show an error after the user presses "Add this contact" if `ContactRequestSend` failed but in this case we also need to obtain ShareableContact in some way and hack/bloat the front architecture to handle the response (the cleanest way to handle it would be to create state in the store that exist only to listen to the result of this call but this departs from the paradigm we're working with currently (only react to events produced by the protocol), we could also add an event on the account log but this is bloat IMO)

glouvigny:
I can provide a call that checks whether the public key and the rdv seed are correctly formatted and returns a `ShareableContact` if it succeed, but you will have to ensure that the app metadata (ie. display name, profile picture) are valid anyway.

`AccountContactRequestReceived` and `AccountContactRequestEnqueued` types can use a `ShareableContact` instead of having 3 distinct fields, same with `ContactRequestSend.Request` which can use the previously unmarshalled data

moul:
@n0izn0iz what's your opinion

n0izn0iz:
I understand that we have to take care of the app metadata

Since ShareableContact doesn't contain it this shouldn't be what is returned IMO, what I propose is to have (maybe find a better name ^^):
```proto
rpc ReferenceToContactPK (ReferenceToContactPK.Request) returns (ReferenceToContactPK.Reply);

message ReferenceToContactPK {
  message Request {
    // reference is an opaque message describing how to connect to the other account
    bytes reference = 1;
  }
  message Reply {
    bytes contact_pk = 1 [(gogoproto.customname) = "ContactPK"];
  }
}
```
Where `bytes reference = 1;` is the same as `types.ContactRequestReference.Reply.reference`

We could also add the rdv seed in the reply but I don't think we need it in the app

glouvigny:
We agreed with @n0izn0iz to use a `ShareableContact` where applicable and remove `reference`s which seem poorly suited to work with in the client.

n0izn0iz:
Closing in favor of #1847

PR #1833, opened 2020-04-06
feat(chat): timestamp and state message in box

### LINKED ISSUE #1735: Put message time and state in message box
We should extend the background of messages to include the message time and state

This is to reduce the gap between messages

The time and state should be displayed in white then

clegirar:
Fixes #1735


n0izn0iz:
Scraped by ux team

PR #1840, opened 2020-04-07
feat(discovery): Override ipfs routing with tinder package

gfanton:
Inject tinder package into ipfs routing system

depends on #1849


go/cmd/berty/main.go
```diff
@@ -42,6 +51,12 @@ import (
 	"moul.io/srand"
 )
 
+// Default ipfs bootstrap & rendezvous point server
+
+const RendezVousServer = "/ip4/167.99.223.55/tcp/4040/p2p/QmTo3RS6Uc8aCS5Cxx8EBHkNCe4C7vKRanbMEboxkA92Cn"
```
moul, line 56:
```suggestion
const DevRendezVousServer = "/ip4/167.99.223.55/tcp/4040/p2p/QmTo3RS6Uc8aCS5Cxx8EBHkNCe4C7vKRanbMEboxkA92Cn"
```

go/cmd/berty/main.go
```diff
@@ -144,12 +140,19 @@ func main() {
...
-				GroupInvitation: *miniClientDemoGroup,
-				Port:            *miniClientDemoPort,
-				RootDS:          rootDS,
-				Logger:          logger,
+				RemoteAddr:            remoteAddr,
+				GroupInvitation:       *miniClientDemoGroup,
+				Port:                  *miniClientDemoPort,
+				RootDS:                rootDS,
+				Logger:                l,
+				Bootstrap:             DefaultBootstrap,
```
moul, line 154: why not using RendezVousServer here too?

go/cmd/berty/main.go
```diff
@@ -169,12 +172,42 @@ func main() {
...
+
+				mardv := multiaddr.StringCast(RendezVousServer)
+				rdvpeer, err := peer.AddrInfoFromP2pAddr(mardv)
+				if err != nil {
+					return errcode.TODO.Wrap(err)
+				}
+
+				routingOpts, crouting := ipfsutil.NewTinderRouting(logger, rdvpeer.ID, false)
+				ipfsOpts := &ipfsutil.IpfsOpts{
+					Bootstrap: append(DefaultBootstrap, RendezVousServer),
```
moul, line 184: It's not clear why sometimes RendezVousServer is a RendezVousServer and why sometimes it's a bootstrap node, will make a better sense to duplicate the const address but use specific variable names

go/cmd/berty/main.go
moul: or DefaultRendezVousServer if you prefer

go/cmd/berty/main.go
```diff
@@ -169,12 +172,42 @@ func main() {
...
 			{
-				api, node, err := ipfsutil.NewInMemoryCoreAPI(ctx)
+
+				mardv := multiaddr.StringCast(RendezVousServer)
+				rdvpeer, err := peer.AddrInfoFromP2pAddr(mardv)
+				if err != nil {
+					return errcode.TODO.Wrap(err)
+				}
+
+				routingOpts, crouting := ipfsutil.NewTinderRouting(logger, rdvpeer.ID, false)
```
moul, line 182: not sure about the naming here, is `routingOpts`, actually a real Opts, or more like a `router`?

is `croutine` a chan to now when it's "done"? if yes, maybe something like `routerFinished`

go/cmd/berty/main.go
```diff
@@ -169,12 +172,42 @@ func main() {
...
+
+				// uncomment this to force connection with rdvp
+				// for {
+				// 	if err := node.PeerHost.Connect(ctx, *rdvpeer); err != nil {
+				// 		logger.Error("cannot dial rendez-vous point: %v", zap.Error(err))
+				// 	} else {
+				// 		break
+				// 	}
+				// 	time.Sleep(time.Second)
+				// }
```
moul, line 206: replace this comment by a CLI flag

go/cmd/berty/main.go
```diff
@@ -434,3 +481,83 @@ func parseAddr(addr string) (maddr ma.Multiaddr, err error) {
 
 	return
 }
+
+func newLogger(debug bool, logfile string) (*zap.Logger, error) {
+	bertyDebug := parseBoolFromEnv("BERTY_DEBUG") || debug
+	libp2pDebug := parseBoolFromEnv("LIBP2P_DEBUG")
+	// @NOTE(gfanton): since orbitdb use `zap.L()`, this will only
+	// replace zap global logger with our logger
+	orbitdbDebug := parseBoolFromEnv("ORBITDB_DEBUG")
```
moul, line 490: use `ff`

go/cmd/berty/main.go
moul, line 56: this should set the default flag

go/internal/ipfsutil/testing.go
```diff
@@ -61,47 +69,92 @@ func TestingRepo(t testing.TB) ipfs_repo.Repo {
...
 }
 
+type TestingAPIOpts struct {
+	Mocknet libp2p_mocknet.Mocknet
+	RDVPeer peer.ID
+}
+
 // TestingCoreAPIUsingMockNet returns a fully initialized mocked Core API with the given mocknet
-func TestingCoreAPIUsingMockNet(ctx context.Context, t testing.TB, m libp2p_mocknet.Mocknet) CoreAPIMock {
+func TestingCoreAPIUsingMockNet(ctx context.Context, t testing.TB, opts *TestingAPIOpts) (CoreAPIMock, func()) {
```
moul, line 78:
```suggestion
func TestingCoreAPIUsingMockNet(ctx context.Context, t testing.TB, opts *TestingAPIOpts) (api CoreAPIMock, cleanup func()) {
```

go/cmd/berty/main.go
gfanton: - it's an actual routing option in ipfs build config: `libp2p.RoutingOption`
- `crouting` output the actual routing used by ipfs (inside a `routingOut` struct) along the dht instance

go/cmd/berty/main.go
gfanton: It's because im not using the rdvp the same way in **mini** and daemon, I don't want to start mini until im fully connected with the rdvp,i gonna change the name and add flags to be more clear

moul: :+1:

PR #1847, opened 2020-04-09
refactor(proto): remove opaque reference and use ShareableContact

n0izn0iz:
opening to discuss the protocol changes, will add the js/go changes here soon


n0izn0iz:
derived code changes done

PR #1850, opened 2020-04-09
fix(chat): check font family

### LINKED ISSUE #1836: Check fonts
Check the main font of the app (Open Sans and Courier)

We need to use ui-kitten Text everywhere and set the ui-kitten default font to the most used font

We also need to fix the `styles.text.family` usage that causes warnings everywhere

For the less used fonts we must use `styles.text.family.use("The-Font")`

clegirar:
Fixes #1836


n0izn0iz:
While checking the fonts in messages I saw that it did not fit mockups too well so I edited a bit

# Before
Screenshot 2020-04-15 at 07 37 38

# After
Screenshot 2020-04-15 at 07 35 08

PR #1896, opened 2020-04-21
Fix ListModal on iOS 12

### LINKED ISSUE #1902: Main modal was broken on iOS 12.3.1
The main modal was broken on many versions of iOS

n0izn0iz:
Fixes #1902


n0izn0iz:
I will merge this since I can't test yolo builds on one of my devices until this is merged and I don't want to upgrade it to iOS 13 so we can catch problems on this version
@clegirar please open an other PR for the swipe behavior

PR #1907, opened 2020-04-22
feat(chat): add mini qr code home settings and navigate

clegirar:
Fixes https://github.com/orgs/berty/projects/48#card-36805343


n0izn0iz:
added in https://github.com/berty/berty/pull/1912

PR #1931, opened 2020-05-07
fix(chat): delete shared props

clegirar:


js/packages/berty-storybook/settings/MyBertyId.tsx
```diff
@@ -84,7 +83,7 @@ const SelectedContent: React.FC<{ contentName: string }> = ({ contentName }) =>
 	}
 }
 
-const BertIdBody: React.FC<RequestProps> = ({ user }) => {
+const BertIdBody: React.FC<{ user: any }> = ({ user }) => {
```
n0izn0iz, line 86: We should try to avoid using `any` during the refactor

PR #1938, opened 2020-05-11
feat(chat): add modal on delete account

clegirar:


n0izn0iz:
Maybe it would be better to add something even more harder to missclick since there is no way of getting your data back
Something like a "type delete in the box below".

PR #1939, opened 2020-05-11
initial contact request manager

glouvigny:


n0izn0iz:
You can merge since it doesn't break the contact request when contacts are added mutually, it dosen't work with only one request though

PR #1948, opened 2020-05-14
fix(chat): refactor styles package

clegirar:


js/packages/styles/types.ts
```diff
@@ -0,0 +1,159 @@
...
+	height: (height: number) => {}
+	maxWidth: (maxWidth: number) => {}
+	maxHeight: (maxHeight: number) => {}
+	opacity: (opacity: number) => {}
+	minWidth: (minWidth: number) => {}
+	minHeight: (minHeight: number) => {}
+	overflow: {}
+}
+
+export const dimensions = Dimensions.get('window')
```
n0izn0iz, line 152: please move those "constants" outside of the types file

PR #1950, opened 2020-05-16
Rework app grpc

n0izn0iz:
mainly wraps improbable-eng/grpc-web's calls in redux-saga's EventChannels to allow for stream cancelation handling and stream end handling

this also:
 - adds the codegen by improbable-eng/grpc-web (ts-protoc-gen)
 - refactors around go-bridge, grpc-bridge and store to better split concerns.
     - everything that knows about bertyprotocol in those packages has been moved to store/protocol
     - interfaces for go-bridge have been moved to it
     - grpc-bridge is now a collection of grpc utilities
- required a bump of protoc from 3.6.1 to 3.11.2 (might be cool with lower but didn't test, 3.6.1 is definitely not fine)
- adds an ack backlog in the store in case we get an ack event before it's target message
- improves the messages box in web-dev-app


.buildkite/pipeline.yml
```diff
@@ -164,9 +164,14 @@ steps:
       queue: "bigcores"
     commands:
       - set -x
-      - cd js
+      - cd go
+      - go mod vendor
```
moul, line 168: a small detail, but this can be a makefile dependency in `js/`

it's not a problem to have this step in buildkite, but it can be a problem to support people that didn't run this command

n0izn0iz: it's already the case locally when you run `make gen` in `js/`

I call the commands directly in the pipeline to avoid runing docker in docker

PR #1964, opened 2020-05-19
feat(chat): add vibrations

clegirar:


zooma2k18:
Onboarding screens confirmations : vibr ok <0,5 stop>
But, no vibr on messages notifications (tested thanks to a quick testsession with moul)

PR #1970, opened 2020-05-20
fix(chat): warnings unique key

clegirar:


js/packages/components/main/List.tsx
```diff
@@ -331,8 +331,8 @@ const Conversations: React.FC<ConversationsProps> = ({ items, hasRequests }) =>
 							>
 								{t('main.messages.title')}
 							</Text>
-							{items.map((_) => {
-								return <ConversationsItem {..._} />
+							{items.map((_, key) => {
```
n0izn0iz, line 334: you should avoid using an array index as key if you can, in this case you can use the conversation id I think

PR #1971, opened 2020-05-20
feat: add an helper to share invite links on a dedicated dev discord channel

moul:
* add a new `berty dev share-invite` command
* add a generic discord-based logging library
* reorganize CLI commands
* add CLI command help messages

### DIFF
--- a/go/cmd/berty/main.go
+++ b/go/cmd/berty/main.go
@@ -138,9 +138,10 @@ func main() {
 	}
 
 	banner := &ffcli.Command{
-		Name:    "banner",
-		Usage:   "banner",
-		FlagSet: bannerFlags,
+		Name:      "banner",
+		Usage:     "banner",
+		FlagSet:   bannerFlags,
+		ShortHelp: "print the ascii Berty banner",
 		Exec: func(args []string) error {
 			cleanup := globalPreRun(context.Background())
 			defer cleanup()
@@ -155,8 +156,9 @@ func main() {
 	}
 
 	version := &ffcli.Command{
-		Name:  "version",
-		Usage: "version",
+		Name:      "version",
+		Usage:     "version",
+		ShortHelp: "print software version",
 		Exec: func(args []string) error {
 			fmt.Println("dev")
 			return nil
@@ -164,9 +166,10 @@ func main() {
 	}
 
 	mini := &ffcli.Command{
-		Name:    "mini",
-		Usage:   "mini",
-		FlagSet: miniClientDemoFlags,
+		Name:      "mini",
+		ShortHelp: "start a terminal-based mini berty client (not fully compatible with the app)",
+		Usage:     "mini",
+		FlagSet:   miniClientDemoFlags,
 		Exec: func(args []string) error {
 			ctx := context.Background()
 			cleanup := globalPreRun(ctx)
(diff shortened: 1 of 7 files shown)


moul:
manfreds-imac - Screen Shot 2020-05-21 at 04 20 56



---

```console
$ berty dev share-invite -d /tmp/berty --discord
```

manfreds-imac - Screen Shot 2020-05-21 at 04 21 10

go/internal/discordlog/discord.go
```diff
@@ -0,0 +1,61 @@
...
+	"berty.tech/berty/v2/go/pkg/errcode"
+	"github.com/itsTurnip/dishooks"
+	qrcode "github.com/skip2/go-qrcode"
+)
+
+// rooms are webhook URLs encoded in base64 to prevent most stupid bots to spam our channels.
+//
+// if we get spammed, then we can think about a better way to store secrets on the repo and in the releases.
+const (
+	QRCodeRoom = "aHR0cHM6Ly9kaXNjb3JkYXBwLmNvbS9hcGkvd2ViaG9va3MvNzEyNjkwNTc5NTk0ODA1Mzc5L2R2OTFFX0pqM0xkYVdFM2p5N0N1TDAtMmRzQ1NTaE85Q2RGRG05QUhtQThCWFRXS0lEZ29jUFRFRDVyTU5KVlFkcFJ5"
```
aeddi, line 16: Why don't we set this in an env variable and/or a flag?
It could be useful to avoid spam and to reuse this package with another room/webhook.

go/cmd/berty/main.go
```diff
@@ -418,11 +422,107 @@ func main() {
 		},
 	}
 
+	shareInvite := &ffcli.Command{
```
aeddi, line 425: Shouldn't be more relevant to add this feature in the app? (Maybe bound to your go package though a gRPC call)
I'm not sure to understand the use case of this command.

go/internal/discordlog/discord_test.go
```diff
@@ -0,0 +1,16 @@
+package discordlog
+
+import (
+	"os"
+	"testing"
+)
+
+func TestShareQRLink(t *testing.T) {
+	if os.Getenv("BERTY_TEST_DISCORD") != "1" {
```
aeddi, line 9: We could replace this env variable by the one containing the QRCodeRoom (if empty -> t.Skip())

go/internal/discordlog/discord.go
moul: > Why don't we set this in an env variable and/or a flag?

to avoid writing runtime doc

> It could be useful to avoid spam

if spam, we reset the webhook url and find a better approach (-> we won't be spammed)

> and to reuse this package with another room/webhook.

yep, that's why I named it QRCodeRoom in a `const(\n)` list of future potential rooms

it's a devtool, make things simple :)

go/cmd/berty/main.go
moul: yes I want to move it to the app, but since it's a new devtool I want to merge this V1 first then talk with @n0izn0iz and @gfanton about how they prefer that we implement this

btw, this command is for sharing your local QR code (the one used with berty mini or a daemon), and it's not limited to sending to discord, actually, it's an opt-in option (cf screenshots)

PR #1975, opened 2020-05-25
chore: bump bazel to 3.1.0

gfanton:
- bump bazel to 3.1.0
- bump go-bridge to go 1.14.3
- update buildkite images

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -107,12 +107,18 @@ steps:
               "HOME=/home/buildkite-agent",
               "GRADLE_USER_HOME=/home/buildkite-agent/.gradle",
             ]
-          volumes: ["~:/home/buildkite-agent"]
+          volumes:
+            [
+              "/usr/local/go:/usr/local/go",
+              "~:/home/buildkite-agent",
+            ]
     agents:
       queue: "bigcores"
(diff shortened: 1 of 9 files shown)


.buildkite/pipeline.yml
```diff
@@ -107,12 +107,18 @@ steps:
...
+              "/usr/local/go:/usr/local/go",
+              "~:/home/buildkite-agent",
+            ]
     agents:
       queue: "bigcores"
     commands:
       - set -x
       - export PATH=$$PATH:/home/buildkite-agent/go/bin
+      - export PATH=/usr/local/go/bin:$$PATH
+      - go version
```
aeddi, line 121: Why?

gfanton: I bump every go version to `1.14.3`, android-build is using a go version prior to 1.14 in `bitriseio/android-ndk:v2019_12_28-08_15-b1793` image, and even the last version of their image (bitriseio/android-ndk) use go 1.13 only, since we didn't have any hand on this, i prefer to use a local version of go so we can still use this image.
Also, I doesn't want to reuse their image to make a new one since the image is more than 10go, and it can be tidy to maintain every time we want to use a new version of go.

PR #1990, opened 2020-06-02
chore: run the 'make generate' CI test to github actions

moul:


moul:
Ok, for now I will only do it for `go/` and `docs/`

`js/` is too slow for an action that should run on each commmit

manfreds-imac - Screen Shot 2020-06-02 at 17 37 45

gfanton:
LGTM 👍

PR #1993, opened 2020-06-02
Share QR code on dev channel (devtool)

n0izn0iz:


n0izn0iz:
Merging altough the format is not supported by the app yet, will add it in another PR

PR #2004, opened 2020-06-02
feat: implement contact request from bertymessenger go package

moul:
This PR implements the full logic of deep links generation and parsing

I also added a placeholder for a future pure-go SendContactRequest which is not yet implemented, but it's already possible to update the front codebase to remove everything related to parsing logic but continue to call the protocol to send the ContactRequest


api/bertymessenger.proto
```diff
@@ -16,6 +16,13 @@ service MessengerService {
   // DevShareInstanceBertyID shares your Berty ID on a dev channel.
   // TODO: remove for public.
   rpc DevShareInstanceBertyID(DevShareInstanceBertyID.Request) returns (DevShareInstanceBertyID.Reply);
+
+  // ParseDeepLink parses a link in the form of berty://xxx or https://berty.tech/id# and returns a structure
+  // that can be used to display information.
+  // This action is read-only.
+  rpc ParseDeepLink(ParseDeepLink.Request) returns (ParseDeepLink.Reply);
+
+  rpc AcceptContactRequest(AcceptContactRequest.Request) returns (AcceptContactRequest.Reply);
```
n0izn0iz, line 25: SendContactRequest would be better IMO since it's what is called underneath

PR #2006, opened 2020-06-04
chore: allow to customize the tracer prefix from the bridge

moul:


moul:
@gfanton, I let you call the setPrefix helper with the dynamic value in another PR

PR #2018, opened 2020-06-05
feat: activate multipeer connectivity by default

D4ryl00:
Signed-off-by: remi <remi@berty.tech>

### DIFF
--- a/go/framework/bertybridge/bridge_protocol.go
+++ b/go/framework/bertybridge/bridge_protocol.go
@@ -9,6 +9,7 @@ import (
 
 	"berty.tech/berty/v2/go/internal/config"
 	"berty.tech/berty/v2/go/internal/ipfsutil"
+	mc "berty.tech/berty/v2/go/internal/multipeer-connectivity-transport"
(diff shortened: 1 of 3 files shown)


go/framework/bertybridge/bridge_protocol.go
```diff
@@ -144,8 +147,10 @@ func newProtocolBridge(logger *zap.Logger, config *ProtocolConfig) (*Protocol, e
 			}
 
 			var bopts = ipfsutil.CoreAPIConfig{
-				APIAddrs:  DefaultAPIAddrs,
-				APIConfig: APIConfig,
+				SwarmAddrs:        []string{defaultMCBind},
```
gfanton, line 150: Can you update the berty config to have `DefaultSwarmAddrs` instead of `DefaultMCBind`, then add all default addrs to it (`tcp` + `quic` + `DefaultMCBind`)
You will need to update all addrs passed by the JS as well.
This will be more readable with less side effect

D4ryl00:
There is a race condition in the gListener. It's not prioritary right now, so I let this PR open for later.

PR #2027, opened 2020-06-07
fix(front): disable autocomplete/capitalization for sensitive fields

### LINKED ISSUE #1996: Handle "Delete " (autocomplete-inserted space) in delete account modal
iOS may insert a space after autocompleting a word, so "Delete " will be entered into the Delete Account modal. The space should be accepted. Or we should disable autocomplete in this field.

Brought up in testing session #2

ekelen:
Blocked by #2022 (since this isn't an urgent issue and it will be much faster to rebase, as 2022 changes yarn.lock)

Closes #1996

### DIFF
--- a/js/packages/components/modals/DeleteAccount.tsx
+++ b/js/packages/components/modals/DeleteAccount.tsx
@@ -97,7 +97,7 @@ function usePrevious<T>(value: T) {
 	return ref.current
 }
 
-const DELETE_STR = 'Delete'
+const DELETE_STR = 'delete'
 
 const DeleteAccountContent: React.FC<{}> = () => {
 	const _styles = useStylesDeleteAccount()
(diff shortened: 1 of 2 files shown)


js/packages/components/modals/DeleteAccount.tsx
```diff
@@ -138,6 +139,8 @@ const DeleteAccountContent: React.FC<{}> = () => {
 					]}
 					value={deleteConfirmation}
 					onChangeText={setDeleteConfirmation}
+					autoCorrect={false}
+					autoCapitalize='none'
```
n0izn0iz, line 143: if we disable `autoCapitalize` we should also set the `DELETE_STR` to 'delete' (without cap) IMO

ekelen: Hah, I agree 100%, I did it in my last push, then set it back. Doing.

PR #2037, opened 2020-06-09
feat(mini): use bertymessenger for contact sharing

glouvigny:


moul:
I can add my terminal as contact from my phone \o/

PR #2039, opened 2020-06-09
feat: improve logger

gfanton:
- Fix log type naming in console (private -> public)
- Forward `console.log` to native logger
- Improve native logger


n0izn0iz:
Can you still display an alert on error pls?
So that non-tech can screen the error during testing sessions

PR #2048, opened 2020-06-10
feat: local discovery

D4ryl00:
Untitled Diagram

Signed-off-by: D4ryl00 <remi@berty.tech>


api/go-internal/localrecord.proto
```diff
@@ -0,0 +1,16 @@
...
+
+option go_package = "berty.tech/berty/go/internal/tinder";
+
+option (gogoproto.goproto_enum_prefix_all) = false;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+
+message LocalRecord {
+  map<string, int64> records = 1;
```
moul, line 15: if possible, avoid maps

D4ryl00: What do you suggest? Two lists, one of keys, and one of value?

moul: the second option is better

D4ryl00: done

go/internal/tinder/driver_localdiscovery.go
```diff
@@ -0,0 +1,394 @@
...
+
+// Delete expired entries of a CID in the peer cache
+func (ld *localDiscovery) cleanPeerCache(cid string) {
+	currentTime := time.Now().Unix()
+
+	ld.peerCacheMux.RLock()
+	_, ok := ld.peerCache[cid]
+	ld.peerCacheMux.RUnlock()
+	if ok {
+		ld.peerCacheMux.Lock()
```
gfanton, line 83:
```suggestion
		ld.peerCacheMux.Lock()
		defer ld.peerCacheMux.Unlock()
```

go/internal/tinder/driver_localdiscovery.go
```diff
@@ -0,0 +1,394 @@
...
+	ld.peerCacheMux.RLock()
+	cache, ok := ld.peerCache[cid]
+	ld.peerCacheMux.RUnlock()
+	if ok {
+		cache.mux.Lock()
+		size = len(cache.recs)
+		cache.mux.Unlock()
+	}
+	return size
+}
```
gfanton, line 128:
```suggestion
func (ld *localDiscovery) peerCacheLen(cid string) (size int) {
	ld.peerCacheMux.RLock()
	if cache, ok := ld.peerCache[cid]; ok {
	        cache.mux.RLock()
		size = len(cache.recs)
	        cache.mux.RUnlock()
	}
	ld.peerCacheMux.RUnlock()
	return
}
```

go/internal/tinder/driver_localdiscovery.go
```diff
@@ -0,0 +1,394 @@
...
+	return ld
+}
+
+// Delete expired entries of a CID in the peer cache
+func (ld *localDiscovery) cleanPeerCache(cid string) {
+	currentTime := time.Now().Unix()
+
+	ld.peerCacheMux.RLock()
+	_, ok := ld.peerCache[cid]
+	ld.peerCacheMux.RUnlock()
```
gfanton, line 81: if you use this patern a lot you should consider using:
```go
func (ld *localDiscovery) GetCache(key string) (ok bool, c pCache) {
     ld.peerCacheMux.RLock()
     ok, c = ld.peerCache[cid]
     ld.peerCacheMux.RUnlock()
     return
}

Comment (D4ryl00) [reply go/internal/tinder/driver_localdiscovery.go]: Done

File: api/go-internal/localrecord.proto [hunk 5]
```diff
@@ -0,0 +1,20 @@
...
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+
+message LocalRecord {
+    message Record {
+      string cid = 1;
+      int64 expire = 2;
+    }
+    repeated Record records = 1;
+}
```
moul, line 20: Maybe it's normal, but it's very strange for me to "one `LocalRecord` that contains many `Records`,

shouldn't it be `LocalRecords` instead?

---

btw, you can get cleaner generated code like this:

```suggestion
message LocalRecord {
    repeated Entry records = 1;
}

message Entry {
  string cid = 1;
  int64 expire = 2;
}
```

the generated code will change from:

```go
 	lr := &LocalRecord{Records: []*LocalRecord_Record{}} 
```

to

```go
 	lr := &LocalRecord{Records: []*Entry{}} 
```

---

buf if things are not voluntarily singular here, I think the best is just:

```suggestion
message Records {
    repeated Record records = 1;
}

message Record {
  string cid = 1;
  int64 expire = 2;
}
```

D4ryl00: ok, I've chosen the second solution to be more generic. Thanks!

PR #2071, opened 2020-06-15
feat(front): connect search functionality to UI

### LINKED ISSUE #2070: Connect basic search function to UI
Related to #2044

ekelen:
! Do not merge. This is in progress and for feedback only.

- [X] Check correct display of search results (highlighting, snippet length)
- [x] Tapping a Contact result will go to conversation with that contact
- [x] (**with bugs**)Tapping a Message result will go to its conversation and scroll to that message 
- [X] Click ➡️ in Header takes you home
- X] Make sure there is no top bounce or unnecessary scrolling enabled
- [X] Clicking 'x' in searchbar OR deleting search manually clears current search and results
- [X] Begin showing results with >= 1 char in search bar
- [x] Contact results with no associated conversation are visibly non-interactive

Known bugs:
- [X] **fixed** Pressing a contact who has not accepted your request will have interaction effect but do nothing
- [X] **fixed** Old search message results being concatenated with new ones, working on it

Blocks #2083 
Blocks #2082 

Closes #2070


js/packages/components/chat/OneToOne.tsx
```diff
@@ -118,14 +119,27 @@ const AppMessage: React.FC<{ message: string }> = ({ message }) => {
 	return msg ? <TypedMessage payload={msg} /> : null
 }
 
-const MessageList: React.FC<{ id: string }> = (props) => {
+const MessageList: React.FC<{ id: string; scrollToMessage?: number }> = (props) => {
 	const [{ row, overflow, flex, margin }] = useStyles()
 	const conversation = Messenger.useGetConversation(props.id)
-	if (!conversation) {
-		return <CenteredActivityIndicator />
+	const flatListRef = useRef<FlatList<any>>(null)
```
clegirar, line 125: You can use `userRef<FlatList<messenger.conversation.Entity>>(null)`

ekelen: Since this is only a list of message IDs, it's `<FlatList<string>>` (added and it works)

ekelen:
Right now the search functions are not very efficient.

e.g. to search a message:
1. Gets a list of all message IDs from all conversations
2. Gets all those messages and searches for `searchText` inside message body, returns that message
3. When rendering this message list, a search will be run through every conversation again to find the conversation the message belongs to

It makes more sense to me to return the conversation metadata with each message so we don't have to do step #3

Help wanted with some solution that doesn't require creating new complicated types :)

PR #2075, opened 2020-06-15
feat(messenger): use URLs for group invites

glouvigny:


api/bertymessenger.proto
```diff
@@ -13,6 +13,9 @@ service MessengerService {
   // InstanceShareableBertyID returns a Berty ID that can be shared as a string, QR code or deep link.
   rpc InstanceShareableBertyID(InstanceShareableBertyID.Request) returns (InstanceShareableBertyID.Reply);
 
+  // InstanceShareableBertyGroup returns a Berty Group that can be shared as a string, QR code or deep link.
+  rpc InstanceShareableBertyGroup(InstanceShareableBertyGroup.Request) returns (InstanceShareableBertyGroup.Reply);
```
moul, line 17: Not sure about the naming here, the first one was about sharing the berty ID of the current instance

go/pkg/bertymessenger/api.go
```diff
@@ -151,13 +140,130 @@ func (s *service) ParseDeepLink(ctx context.Context, req *ParseDeepLink_Request)
...
+	return &ret, nil
+}
+
+func (s *service) InstanceShareableBertyGroup(ctx context.Context, request *InstanceShareableBertyGroup_Request) (*InstanceShareableBertyGroup_Reply, error) {
+	grpInfo, err := s.protocol.GroupInfo(ctx, &bertytypes.GroupInfo_Request{
+		GroupPK:              request.GroupPK,
+		ContactPK:            nil,
+		XXX_NoUnkeyedLiteral: struct{}{},
+		XXX_unrecognized:     nil,
+		XXX_sizecache:        0,
```
moul, line 217: can this be removed?

moul:
> Should we deprecate `bertyprotocol.MultiMemberGroupInvitationCreate`?

To take this decision, we need to start maintaining a tested, simple hello world app in the monorepo, that uses bertyprotocol without bertymessenger

api/bertymessenger.proto
glouvigny: Ok, renaming to `ShareableBertyGroup`

go/pkg/bertymessenger/api.go
glouvigny: My bad, thank you

glouvigny:
> To take this decision, we need to start maintaining a tested, simple hello world app in the monorepo, that uses bertyprotocol without bertymessenger

Ok, no need to rush. I'll keep it as is for now.

PR #2084, opened 2020-06-17
feat: Add grpc gateway

gfanton:
Add grpc gateway with`grpcgw` protocol
```sh
❯  berty daemon -l /ip4/127.0.0.1/tcp/9091/grpcgw
```


moul:
Manfreds-iMac - Screen Shot 2020-06-18 at 02 46 47

moul:
@gfanton, I fixed most of the build issues but I can't fix iOS and Android builds easily

I think that we should make a copy of the .proto files and strip the annotations stuff before generating the JS libs

cc @n0izn0iz

PR #2087, opened 2020-06-18
chore: Improve CI logging

gfanton:


.buildkite/pipeline.yml
```diff
@@ -114,5 +117,23 @@ steps:
...
+      -  junit/$BUILDKITE_COMMIT/*.xml
+
+  - wait:
+    continue_on_failure: true
+  - label: reports-bigcors
+    agents:
+      queue: "bigcores"
+    plugins:
+      - junit-annotate#v1.7.0:
+          artifacts: junit/$BUILDKITE_COMMIT/*.xml
```
n0izn0iz, line 139: This won't work
Each job is sent to a different agent and the repo is cleared at the start of the job

n0izn0iz: My bad ^^ I didn't understand that it uses the artifacts of the other jobs

PR #2092, opened 2020-06-19
feat(mini): defaulting to a persistent store + support for display name as a cli arg

glouvigny:


go/cmd/berty/main.go
```diff
@@ -208,6 +211,20 @@ func main() {
 			cleanup := globalPreRun()
 			defer cleanup()
 
+			if !miniInMemory && datastorePath == cacheleveldown.InMemoryDirectory {
+				storagePath := configdir.New("Berty Technologies", "Berty")
```
moul, line 215: Manfreds-iMac - Screen Shot 2020-06-19 at 18 01 01

it works but looks very strange on linux, I'm okay to merge like this but I prefer to let you choose if you want to make it differently or not

moul:
I tried to rebase master by myself but I can't push on your fork (I can on @gfanton's one)

PR #2094, opened 2020-06-19
feat: update ipfs/libp2p

gfanton:


go/internal/ipfsutil/http_api.go
```diff
@@ -1,9 +1,6 @@
 package ipfsutil
 
 import (
-	"net/http"
```
moul, line 4: to be rollbacked before merging

PR #2102, opened 2020-06-22
WIP chore: setup report annotations

moul:


go/pkg/bertyprotocol/testing_test.go
```diff
@@ -37,3 +20,36 @@ func testSameErrcodes(t *testing.T, expected, got error) {
...
+	require.NoError(t, err)
+}
+
+func TestTestingProtocol(t *testing.T) {
+	ctx := context.Background()
+	opts := TestingOpts{}
+	tp, cleanup := NewTestingProtocol(ctx, t, &opts)
+	assert.NotNil(t, tp)
+	cleanup()
+}
```
n0izn0iz, line 43: It would be nice to test for goroutines leaks now that we need to support restarting the node services (maybe it's already the case)

moul:
Depends on https://github.com/report-ci/scripts/issues/17

go/pkg/bertyprotocol/testing_test.go
moul: I started a PR in go-orbit-db -> https://github.com/berty/go-orbit-db/pull/42

PR #2128, opened 2020-06-29
fix: allow incoming requests from contacts without rdv seed

glouvigny:


go/pkg/bertyprotocol/store_metadata.go
```diff
@@ -565,7 +565,7 @@ func (m *metadataStore) ContactRequestIncomingReceived(ctx context.Context, cont
 		return nil, errcode.ErrGroupInvalidType
 	}
 
-	if err := contact.CheckFormat(); err != nil {
+	if err := contact.CheckFormat(bertytypes.ShareableContactOptionsAllowMissingRDVSeed); err != nil {
```
moul, line 568: Am I missing something?

How before and after can both compile?

moul: Ok: `go/pkg/bertytypes/contact.go:func (m *ShareableContact) CheckFormat(options ...ShareableContactOptions) error {`

go/pkg/bertyprotocol/contact_request_manager_test.go
```diff
@@ -134,3 +134,89 @@ func TestContactRequestFlow(t *testing.T) {
...
+	ref0, err := pts[0].Client.ContactRequestResetReference(ctx, &bertytypes.ContactRequestResetReference_Request{})
+	require.NoError(t, err)
+	require.NotNil(t, ref0)
+
+	subCtx, subCancel := context.WithCancel(ctx)
+	defer subCancel()
+
+	subMeta0, err := pts[0].Client.GroupMetadataSubscribe(subCtx, &bertytypes.GroupMetadataSubscribe_Request{
+		GroupPK: config0.AccountGroupPK,
+		Since:   []byte("give me everything"),
```
moul, line 174: :)

PR #2160, opened 2020-07-16
test: improve scenario tests

aeddi:
Added generic helpers that:
- Create a MultiMember Group
- Send messages on all kind of groups
- Add accounts as contacts

This way we will be able to mix the different elements much more easily to create new scenarios.

Also made generic the function to launch a scenario and its test cases.

### DIFF
--- a/go/pkg/bertyprotocol/scenario_test.go
+++ b/go/pkg/bertyprotocol/scenario_test.go
@@ -1,50 +1,280 @@
 package bertyprotocol
 
 import (
+	"bytes"
 	"context"
+	"encoding/base64"
 	"fmt"
 	"io"
+	"runtime"
+	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
-	crand "crypto/rand"
-
-	"sync"
-
-	"bytes"
-
 	"berty.tech/berty/v2/go/internal/testutil"
 	"berty.tech/berty/v2/go/internal/tracer"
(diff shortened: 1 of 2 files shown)


go/pkg/bertyprotocol/scenario_test.go
```diff
@@ -220,242 +383,442 @@ func testingScenario_JoinGroup(ctx context.Context, t *testing.T, pts ...*Testin
...
-	}
+					// Check if received message was expected
+					expectedMessagesLock.Lock()
+					_, expected := expectedMessages[string(res.Message)]
+					expectedMessagesLock.Unlock()
+					if !expected {
+						continue
+					}
 
-	testutil.SkipUnstable(t) // after this line, the test is unstable, especially on low hardware
```
moul, line 397: you should keep this

aeddi: No need to keep it anymore since the function has been split, before it was useful to skip the last part of a monolithic function that was creating accounts -> adding them as contacts -> sending/receiving messages.
Now this function only sends messages so the skip can be done upstream, on a case by case basis and in a more atomic way.

moul: So, now, the CI with always be green or randomly be green?

aeddi: We can have it less randomly green while having more accurate tests depending on the parts / sub-parts of the tests we chose to skip. :)

Issue #1852, opened 2020-04-09
Integrate new main modal


clegirar:
Integrate the main modal of David (freelance)

moul:
old, closing

PR #2169, opened 2020-07-17
fix(messenger): fix main modal and QR scanner sizing

ekelen:
Just quick fixes to remove blocking responsiveness issues (i.e. can't access some interactive UI element).

- QR scanner now always fits on screen
- Main modal is dismissable and interactive on short or landscape screen
- QR scan page is scrollable for access to manual deeplink add form

Related to #1852

### DIFF
--- a/js/packages/components/main/HomeModal.tsx
+++ b/js/packages/components/main/HomeModal.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react-native/no-inline-styles */
 import React from 'react'
 import {
 	View,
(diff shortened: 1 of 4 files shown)


moul:
> QR scanner now always fits on screen

yes it fits, but we are about 10px missing to be able to read the text below on ipad :)

moul:
I've added on discord (on e-meeting-note) a quick video about how it feels on iPad

moul:
btw, except those bugs, the experience is much greater :)

ekelen:
I'm going to add large screen support for this, as there are other situations where we need it. (basically anywhere that adds conditional scroll on landscape – scroll function should be removed if it's landscape on large screen).

PR #2172, opened 2020-07-20
feat(messenger): add timestamp to message search result

ekelen:
- Adds timestamps to message search result list
- Removes sticky section headers (identified in last testing session), so Contacts section header will disappear if you scroll past that section


zxxma:
LGTM 👍

PR #2188, opened 2020-07-22
feat(messenger): add avatar and name to multimember conversation

### LINKED ISSUE #2187: Add avatars to multimember conversations

ekelen:
Credit for this goes to 
- Adds avatar and username (already done by @clegirar or @n0izn0iz, I've just added it back)
- Puts username inside message body container
- Hides username, timestamp and avatar on follow-up messages (messages where previous message is from same non-self user and is more than 10 minutes time difference)

image

Closes #2187


zxxma:
LGTM! 👍

PR #2190, opened 2020-07-23
chore(messenger): replace inline flex styles

ekelen:
Replace inline flex styles with flex defs from styles package.


ekelen:
Decided against this, as we have no reason (other than syntax and maybe some small performance issues(?)) to manually add primitive css styles to our Styles API.

PR #2194, opened 2020-07-24
fix: use fixed versions for protoc tools

moul:
also make smaller docker image


tool/docker-protoc/Makefile
```diff
@@ -1,10 +1,9 @@
 IMAGE ?=	bertytech/protoc
-VERSION ?=	24
+VERSION ?=	25
 
 build:
-	docker build -t $(IMAGE):$(VERSION) .
+	cd ../../ && docker build -f ./tool/docker-protoc/Dockerfile -t $(IMAGE):$(VERSION) -t $(IMAGE):latest .
```
moul, line 5: sorry for this hack, I don't like it, but I don't find a better way to do it

PR #2204, opened 2020-08-06
feat(messenger): instabug

clegirar:


aeddi:
To use the environment system, you must create a `js/.env` file.

Right now, only the following two variables are useful:
- `INSTABUG_TOKEN=12345` -> will be used to set the token in Java (Android) and JS code
- `INSTABUG_SKIP_UPLOAD=true` -> will be used on Android to skip the Instabug sourcemap upload

Note that on Android / Gradle, variables from the system environment have priority over those from the .env file.

PR #2208, opened 2020-08-10
continue moving the app's logic to golang

n0izn0iz:
- fixed db ops
- added react context that encapsulate the messenger state and api (used in web-dev-app for now)
- added "contact and interaction"-related models, handlers and calls
- added 1to1 test
- added multimember test
- added "conversation with invitations to contacts" test


go/pkg/bertymessenger/debug.log
```diff
@@ -0,0 +1,2808 @@
+2020-08-17T04:41:48.029+0200	[31mERROR[0m	bertymessenger/testing.go:57	trace	{"error": "record not found", "elapsed": "145.901µs", "rows": 0, "sql": "SELECT * FROM `accounts` ORDER BY `accounts`.`public_key` LIMIT 1"}
```
moul, line 1: should be gitignored

go/pkg/bertymessenger/service_test.go
```diff
@@ -119,12 +119,828 @@ func TestContactRequest(t *testing.T) {
...
+					require.False(t, inte.GetIsMe())
+					gotAck = true
+				}
+			}
+		}
+	}
+}
+
+func requireNoEOF(t *testing.T, l *zap.Logger, err error, format string, a ...interface{}) {
+	if err == io.EOF {
```
moul, line 481:
```suggestion
	t.Helper()
	if err == io.EOF {
```

PR #2212, opened 2020-08-14
chore: adding unit tests for cryptoutil

gobins:
Adding basic unit test for cryptutil package

### DIFF
--- a/go/internal/cryptoutil/cryptoutil_test.go
+++ b/go/internal/cryptoutil/cryptoutil_test.go
@@ -1,0 +1,49 @@
+package cryptoutil
+
+import (
+	"crypto/rand"
+	"testing"
+
+	"berty.tech/berty/v2/go/pkg/errcode"
+	"github.com/libp2p/go-libp2p-core/crypto"
+)
+
+func TestSeedFromEd25519PrivateKey(t *testing.T) {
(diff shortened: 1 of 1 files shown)


go/internal/cryptoutil/cryptoutil_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+
+	"github.com/libp2p/go-libp2p-core/crypto"
+)
+
+func TestSeedFromEd25519PrivateKey(t *testing.T) {
+	priv, _, _ := crypto.GenerateECDSAKeyPair(rand.Reader)
+	_, err := SeedFromEd25519PrivateKey(priv)
+	if err == nil {
+		t.Error("Shoulf fail with ErrInvalidInput")
```
moul, line 14:
```suggestion
		t.Error("Should fail with ErrInvalidInput")
```

btw, you can use our helpers from https://pkg.go.dev/berty.tech/berty/v2/go/pkg/errcode?tab=doc#ErrCode.Code to compare the code instead of just checking if err is nil

gobins: Thanks for the review I have updated the PR.

PR #2223, opened 2020-08-20
feat(messenger): crashlytics and shakebugs on dev release

clegirar:


aeddi:
For information: we set up Crashlytics to have something functional quickly. It will only be used during the closed beta and will be removed or replaced by an opt-in solution, more privacy friendly and if possible self-hosted.

PR #2242, opened 2020-08-27
feat: add logs to db replay in messenger

aeddi:


go/pkg/bertymessenger/util.go
```diff
@@ -2,17 +2,23 @@ package bertymessenger
...
 	"bytes"
+	"context"
 	"encoding/base64"
 
+	"berty.tech/berty/v2/go/pkg/bertyprotocol"
 	"berty.tech/berty/v2/go/pkg/bertytypes"
 	"berty.tech/berty/v2/go/pkg/errcode"
 )
 
 func (svc *service) checkIsMe(gme *bertytypes.GroupMessageEvent) (bool, error) {
```
moul, line 13: ambiguous, can you keep only one?

go/pkg/bertymessenger/util.go
```diff
@@ -24,8 +30,12 @@ func (svc *service) checkIsMe(gme *bertytypes.GroupMessageEvent) (bool, error) {
 }
 
 func (svc *service) groupPKFromContactPK(contactPK []byte) ([]byte, error) {
```
moul, line 32: same here, keep only one

moul (approved):
except for the two small details, I really like the way it's done :)

go/pkg/bertymessenger/util.go
aeddi: Yes I hesitated while doing it, I will change it ;)

go/pkg/bertymessenger/util.go
aeddi: 👍

PR #2253, opened 2020-08-30
messenger: restore members

n0izn0iz:
- add members and devices in go store
- add members in js faker
- add member names test
- restore members ui
- fix various bugs in messenger

Preview (don't mind broken android ui):
Screenshot from 2020-08-30 12-54-26


go/pkg/bertymessenger/api.go
```diff
@@ -448,6 +448,25 @@ func (svc *service) EventStream(req *EventStream_Request, sub MessengerService_E
 		}
 	}
 
+	// send members
```
moul, line 451: send after conversations

go/pkg/bertymessenger/handlers.go
```diff
@@ -326,18 +338,91 @@ func groupMemberDeviceAdded(svc *service, gme *bertytypes.GroupMetadataEvent) er
...
+		OwnerPublicKey: mpk,
+	}
+	{
+		err := svc.db.Where(&Device{PublicKey: dpk}).Take(&Device{}).Error
+		if err == gorm.ErrRecordNotFound {
+			err = svc.db.
+				Clauses(clause.OnConflict{
+					Columns:   []clause.Column{{Name: "public_key"}},
+					DoNothing: true,
+				}).
```
moul, line 375:
```suggestion
```

go/pkg/bertymessenger/handlers.go
```diff
@@ -435,76 +520,154 @@ func handleAppMessage(svc *service, gpk string, gme *bertytypes.GroupMessageEven
...
-		svc.logger.Warn("unsupported app message type", zap.String("type", amt.String()))
+		return nil
+	}
+
+	if isMe || conv.GetType() == Conversation_ContactType {
+		return next(nil)
+	}
+
+	existingDevice := &Device{}
+	err = svc.db.Where(&Device{PublicKey: dpk}).Take(existingDevice).Error
```
moul, line 642:
```suggestion
	existingDevice := &Device{PublicKey: dpk}
	err = svc.db.Where(existingDevice).Take(existingDevice).Error
```

I prefer writing this way, don't know about you

go/pkg/bertymessenger/service_test.go
```diff
@@ -432,18 +433,24 @@ func Test3PeersCreateJoinConversation(t *testing.T) {
...
 	{
-		for i := 0; i < 2; i++ {
+		for i := 0; i < joinersAmount; i++ {
 			joiners[i] = NewTestingAccount(ctx, t, clients[i+1], logger)
 			defer joiners[i].Close()
+			accounts[i+1] = joiners[i]
 			joiners[i].DrainInitEvents(t)
+			joiners[i].SetNameAndDrainUpdate(t, "Joiner #"+strconv.Itoa(i))
 		}
 	}
```
moul, line 456: you can simplify this init and make it more readable by keeping the old code and just do this at the end:

```go
accounts := append(joiners, creator)
```

moul (approved):

go/pkg/bertymessenger/handlers.go
n0izn0iz: I prefer my way because it make explicit that the object passed to `.Where(` is only a source

Issue #2227, opened 2020-08-21
[META] current focus: opening the beta 🚀


n0izn0iz:
We should create a branch with a clean store and be based on that or merge it (but this would break the app temporarily)
Base branch merged (https://github.com/berty/berty/pull/2228)

We should also try to split the work as to minimize conflicts

## TODO Go:
- [x] Replay
- [x] Read/Unread logic
- [x] Requests and conversations state dates (https://github.com/berty/berty/pull/2276)
- [x] Conversations sort date
- [x] MultiMember members (https://github.com/berty/berty/pull/2253)
- [ ] Delete and discard calls
- [x] Persistence bug

## TODO JS:
- [x] 1) account views (procedural avatar, fingerprint, display name)
- [x] 2) contact requests/accept (https://github.com/berty/berty/pull/2235)
   - [x] 2.1) scan screen
   - [x] 2.2) outgoing requests
   - [x] 2.3) incoming requests
   - [x] 2.4) link screen
- [x] 3) create multimember (https://github.com/berty/berty/pull/2237)
- [x] 4) contact group invitation (depends on 2, 3, 8, 9.1)
- [x] 5) search (https://github.com/berty/berty/pull/2278)
   - [x] 5.1) basic contact search (https://github.com/berty/berty/pull/2235)
   - [x] 5.2) contact search first message (depends on 9.1 or 7.3)
   - [x] 5.3) message search (depends on 9.1, 9.2 or 7.3)
- [x] 6) unread count/read update (depends on implementing it in the go store and on 8, 9) (https://github.com/berty/berty/pull/2271)
- [x] 7) devtools
  - [x] 7.1) discord share (https://github.com/berty/berty/pull/2235)
  - [x] 7.2) store dumps (https://github.com/berty/berty/pull/2235)
  - [x] 7.3) faker (https://github.com/berty/berty/pull/2234)
  - [x] 7.4) config toggles
  - [x] 7.5) system info (https://github.com/berty/berty/pull/2232)
  - [x] 7.6) send messages (https://github.com/berty/berty/pull/2254)
- [X] 8) conversations list (depends on 2, 3)
- [x] 9) interactions list (depends on 8)
   - [x] 9.1) 1to1
   - [x] 9.2) multimember
- [x] 10) update js/readme
- [x] 11) multimember members (https://github.com/berty/berty/pull/2253)
- [x] 12) conversations sorting
- [x] 13) contact details (depends on 8)
- [x] 14) outgoing request state ui
[...]

PR #2254, opened 2020-08-31
chore(messenger): restore send to all devtool

ekelen:
Restores send test message to all devtool (messenger + webdev app)

Relates to #2227


n0izn0iz (approved):
lgtm

PR #2262, opened 2020-09-01
fix: Panic in WaitGroup in the Multipeer Connectivity transport

D4ryl00:
bug: negative counter in the **WaitGroup** field of the `Listener` struct.

In outgoing connections (`Dial()` method), no **Add** method of WaitGroup is called. So when `newConn` calls **Done** of WaitGroup, the counter becomes negative.

### DIFF
--- a/go/cmd/berty/mini/main.go
+++ b/go/cmd/berty/mini/main.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


gfanton (approved):
LGTM 👍

aeddi (approved):
LGTM 👍

PR #2276, opened 2020-09-06
messenger: restore state dates (created, sent)

n0izn0iz:
Signed-off-by: Norman Meier <norman@berty.tech>


docs/messenger/api.md
```diff
@@ -239,6 +239,8 @@
 | conversation | [Conversation](#berty.messenger.v1.Conversation) |  |  |
 | state | [Contact.State](#berty.messenger.v1.Contact.State) |  |  |
 | display_name | [string](#string) |  |  |
+| created_date | [int64](#int64) |  |  |
+| sent_date | [int64](#int64) |  | specific to outgoing requests |
```
Jorropo, line 243: What is the difference beetween `created_date` and `sent_date` ? Is there a case where we create a request and forward it later ?

n0izn0iz: In the outgoing request case, if the contact we want to request is not online, the contact will stay in `OutgoingRequestEnqueued` state until the contact receives the request.
`created_date` is the date at which we received the `OutgoingRequestEnqueued` or `IncomingRequest` event.
`sent_date` is the date at which we received the `OutgoingRequestSent` event.

Jorropo: :ok_hand:

Jorropo (approved):
LGTM

PR #2281, opened 2020-09-07
feat: added token issuer server for upcoming services

glouvigny:
## Token server

A command to start a token server has been added. The server don't do much expect signing values.

Example command:

```
berty token-server -secret IEKjLRbLY8Oe0G4RcVMCkJx7QNhAR0QFHWpzisX0PJC -sk IEKjLRbLY8Oe0G4RcVMCkJx7QNhAR0QFHWpzisX0PJC -s rpl@127.0.0.1:9091
-secret  is the symmetric secret shared among services
-sk is the private key signing tokens
-s is the comma separated list of services that will be announced to the client, ie. here we announce a rpl service running on 127.0.0.1:9091
```

### Service authentication, discovery

* `/services auth init` : Inits authentication with a service provider
	This commands need a server url as argument. It will output an URL.
* `/services auth complete` : Completes authentication with a service provider
	This commands need a the redirected token URL ie. berty://services-auth/?code=eyJhbGc...&state=blah (on desktop it will be displayed on a web browser as a link but you can copy/paste it).
	On success the token identifier and granted services will be displayed.
* `/services list` : Lists registered services`
	Registered tokens are identified with a UUID and shown.


aeddi (approved):
LGTM 👍

PR #2287, opened 2020-09-08
feat: Add notification driver

gfanton:
- Add notification manager to messenger to trigger local notification on desktop & mobile (can be noop)
- Add Swift notification manager driver
- set minimal ios version to 12.0
- move bundler to js root folder add cocoapods as a gem depencies


go/internal/notification/manager_logger.go
```diff
@@ -0,0 +1,31 @@
...
+		zap.String("title", notif.Title),
+		zap.String("body", notif.Body))
+	return nil
+}
+
+func (m *LoggerManager) Schedule(notif *Notification, interval time.Duration) error {
+	m.logger.Info("scheduling notification", zap.Duration("interval", interval))
+	time.AfterFunc(interval, func() { _ = m.Notify(notif) })
+	return nil
+}
```
moul, line 31: should be async, nope?

go/internal/notification/manager_desktop.go
```diff
@@ -0,0 +1,35 @@
...
+
+// DesktopManager is a Manager
+var _ Manager = (*DesktopManager)(nil)
+
+type DesktopManager struct {
+	logger  *zap.Logger
+	appicon string
+}
+
+func NewDesktopManger(logger *zap.Logger, appicon string) Manager {
```
moul, line 18:
```suggestion
func NewDesktopManager(logger *zap.Logger, appicon string) Manager {
```

go/internal/notification/desktop.go
```diff
@@ -0,0 +1 @@
+package notification
```
moul, line 1:
```suggestion
```

moul (approved):
LGTM

PR #2289, opened 2020-09-08
fix: race condition in the MultipeerConnectivity transporter

D4ryl00:
The global variable `gListener` is not protected by concurrent execution.
For now, this transporter supports only once instance.


Jorropo (commented):
I think you are missing a lock, if 2 listen are issued at the same time an Once will not prevent the race, you might return nil on the second one while the first one is creating it.

Jorropo (approved):
LGTM, For that little it might not be worth an RWMutex (the extra cost of the RWMutex isn't worth it for some struct unpack (the only time it could be worth is for the `gListener.inUse.Wait()` but too bad it's an Lock here)).

aeddi (approved):
LGTM 👍

PR #2291, opened 2020-09-08
fix(messenger): create group screen

clegirar:
On the first screen of the create group, we have some UI issues:
- We have not a size limit for the name of the member's list
- The scroll view scroll only when we scroll on the checkbox (right of the screen)


zxxma:
LGTM 👍

PR #2292, opened 2020-09-09
chore(messenger): restore group invitation functionality in 1-1 chat

ekelen:
- restore group invitation functionality in 1-1 chat

**also**:

- fix(messenger): add missing break stmt to app reducer
- refactor(messenger): add useClient and useOneToOneContact hooks
- fix(messenger): fix encode search and replace

Related to #2227


js/packages/store/provider.tsx
```diff
@@ -46,8 +46,10 @@ const reducer = (oldState: any, action: { type: string; payload?: any }) => {
 				state.members[member.conversationPublicKey] = {}
 			}
 			state.members[member.conversationPublicKey][member.publicKey] = member
+			break
```
n0izn0iz, line 49: nice catch

PR #2306, opened 2020-09-13
fix: multipeer transport race condition

phanhuynhquy:
Signed-off-by: phanquy <phanhuynhquy@gmail.com>

### [Change Description]
Some improve in HandleFoundPeer()
- Missing mutex and checking valid of gListener in go routine forked in HandleFoundPeer() - so if Listener.Close() is called, potentially gListener is set is nill before go routine is executed
- mutex cover also  the case `case listener.inboundConnReq <- connReq{` - this can be blocked in case missing call to Accept(), and will block Close(), Dial()...

### [Test]
None

### DIFF
--- a/go/framework/bertybridge/bridge_test.go
+++ b/go/framework/bertybridge/bridge_test.go
@@ -121,8 +121,6 @@ func TestPersistenceProtocol(t *testing.T) {
 
 	const n_try = 4
 
-	testutil.FilterStabilityAndSpeed(t, testutil.Unstable, testutil.Slow)
-
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/go/internal/multipeer-connectivity-transport/discovery.go
+++ b/go/internal/multipeer-connectivity-transport/discovery.go
@@ -25,22 +25,30 @@ func HandleFoundPeer(sRemotePID string) bool {
 
 	// Checks if a listener is currently running.
 	gLock.RLock()
-	defer gLock.RUnlock()
+
 	if gListener == nil || gListener.ctx.Err() != nil {
+		gLock.RUnlock()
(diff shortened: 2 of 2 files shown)


go/framework/bertybridge/bridge_test.go
```diff
@@ -121,8 +121,6 @@ func TestPersistenceProtocol(t *testing.T) {
 
 	const n_try = 4
 
-	testutil.FilterStabilityAndSpeed(t, testutil.Unstable, testutil.Slow)
```
phanhuynhquy, line 124: This test was failed due to race condition happen in multipeer-connectivity , so we can remove it. But let wait the test run on server.

go/internal/multipeer-connectivity-transport/discovery.go
```diff
@@ -25,22 +25,30 @@ func HandleFoundPeer(sRemotePID string) bool {
...
 
 	// Peer with lexicographical smallest peerID inits libp2p connection.
-	if gListener.Addr().String() < sRemotePID {
+	if listener.Addr().String() < sRemotePID {
 		// Async connect so HandleFoundPeer can return and unlock the native driver.
 		// Needed to read and write during the connect handshake.
 		go func() {
-			err := gListener.transport.host.Connect(gListener.ctx, peer.AddrInfo{
+			// Need to use listener than gListener here to not have to check valid value of gListener
+			err := listener.transport.host.Connect(listener.ctx, peer.AddrInfo{
```
phanhuynhquy, line 51: missing gMutex here, and check valid of gListener.
But the call `Listener.transport.host.Connect()` might not executed inside gLock mutex, maybe inside this function can be call Transport.Dial() where gLock.RLock() is called. I'm not total sure here. need to check source code of libp2p.

So that why snapshot is good option here.

moul (approved):

PR #2307, opened 2020-09-13
feat(messenger): add and display sentDate in GroupInvitation messages

ekelen:


ekelen:
@n0izn0iz / @clegirar If this isn't the correct way to add sentDate to the API, I can remove it from this PR and keep changes in the front only.

n0izn0iz (approved):
lgtm

PR #2309, opened 2020-09-14
chore: Added gitignore to exclude go binaries.

Jorropo:


moul (commented):
Can you move those checks in the top-level .gitignore file to avoid creating those new files?

Jorropo:
> Can you move those checks in the top-level .gitignore file to avoid creating those new files?

@moul done :)

moul (approved):

PR #2310, opened 2020-09-14
feat(messenger): notifications first pass

n0izn0iz:


clegirar (approved):
LGTM (js) 👍

api/bertymessenger.proto
```diff
@@ -386,6 +387,21 @@ message StreamEvent {
     Device device = 1;
   }
   message ListEnd {}
+  message Notified {
+    Type type = 1;
+    string title = 3;
+    string body = 4;
+    bytes payload = 5;
+    enum Type {
+      TypeBasic = 0;
```
moul, line 396: always set a "unknown" value for the first (0) entry

```suggestion
      TypeUnknown = 0;
      TypeBasic = 1;
```

go/pkg/bertymessenger/db.go
```diff
@@ -124,6 +124,18 @@ func getInteraction(db *gorm.DB, cid string) (Interaction, error) {
 	return interactions[0], nil
 }
 
+func getMember(db *gorm.DB, publicKey string) (Member, error) {
```
moul, line 127:
```suggestion
func getMemberByPubkey(db *gorm.DB, publicKey string) (Member, error) {
```

go/pkg/bertymessenger/db.go
```diff
@@ -124,6 +124,18 @@ func getInteraction(db *gorm.DB, cid string) (Interaction, error) {
...
+	var members []Member
+	err := db.Where(&Member{PublicKey: publicKey}).Find(&members).Error
+	if err != nil {
+		return Member{}, err
+	}
+	if len(members) == 0 {
```
moul, line 133: very good way to avoid having an error in the logs :+1: 

however, can you add a check that "should" never happen: `if len(members) > 1` so you can raise an error instead of ignoring this strange thing?

moul (approved):
except for the small styling suggestions, LGTM

moul (approved):

PR #2312, opened 2020-09-14
feat: add a message cache for messages that can't be opened

D4ryl00:
**Scenario:** when A sends to B a message and B has not registered the A's chainkey yet, so B can't decrypt A's message.
**Solution:** we add this message to a cache until the A´s chainkey is registered by B.

Require sending chainkey before sending the first message in a contact group is now in https://github.com/berty/berty/pull/2447


go/pkg/bertyprotocol/store_message.go
```diff
@@ -40,7 +44,80 @@ func (m *messageStore) setLogger(l *zap.Logger) {
...
+}
+
+// Add the event into a slice of the map
+func (m *messageStore) addToCache(devicePK []byte, e ipfslog.Entry) {
+	m.logger.Debug("addToCache", zap.Any("devicePK", devicePK), zap.Any("event", e))
+	m.cacheLock.Lock()
+	m.cache[string(devicePK)] = append(m.cache[string(devicePK)], e)
+	m.cacheLock.Unlock()
+
+	if !m.queueScanCache(devicePK) {
```
D4ryl00, line 63: addToCache must try to open messages in the cache because sometimes, a message is being added into the cache while in another goroutine, the chainkey is registered before this message has finished being added into the cache (so the cache would be empty and the message that will be added would never be opened).

api/bertyprotocol.proto
```diff
@@ -42,6 +42,9 @@ service ProtocolService {
   // ContactUnblock unblocks a contact from sending requests
   rpc ContactUnblock (types.v1.ContactUnblock.Request) returns (types.v1.ContactUnblock.Reply);
 
+  // WaitingForContactReady return a stream saying if the contact is ready to receive messages
+  rpc WaitingForContactReady (types.v1.WaitingForContactReady.Request) returns (stream types.v1.WaitingForContactReady.Reply);
```
moul, line 46:
```suggestion
  // WaitContactIsReady returns a stream that will send an event as soon as a contact is ready to receive messages.
  rpc WaitContactIsReady (types.v1.WaitingForContactReady.Request) returns (stream types.v1.WaitingForContactReady.Reply);
```

btw, is it really "contact", not "member" or "device" or anything else?

api/bertytypes.proto
```diff
@@ -565,6 +565,17 @@ message ContactUnblock {
   message Reply {}
 }
 
+message WaitingForContactReady {
+  message Request {
+    // group_pk is the identifier of the group
+    bytes group_pk = 1 [(gogoproto.customname) = "GroupPK"];
+    // device_pk is the member's device
+    bytes device_pk = 2 [(gogoproto.customname) = "ContactPK"];
```
moul, line 573: is it `device_pk` or `ContactPK`?

go/pkg/bertyprotocol/store_metadata.go
```diff
@@ -286,6 +286,31 @@ func metadataStoreAddEvent(ctx context.Context, m *metadataStore, g *bertytypes.
 	return op, nil
 }
 
+// ContactReady informs if secrets are sent to the member, so if it can read messages send by us
```
moul, line 289:
```suggestion
// ContactReady waits until the secrets are sent to the member.
```


not sure whether it's "Contact" or "Member"

go/pkg/bertyprotocol/store_metadata.go
```diff
@@ -286,6 +286,31 @@ func metadataStoreAddEvent(ctx context.Context, m *metadataStore, g *bertytypes.
...
+func (m *metadataStore) ContactReady(ctx context.Context, memberPK crypto.PubKey) (bool, error) {
```
moul, line 290:
```suggestion
func (m *metadataStore) ContactReady(ctx context.Context, memberPK crypto.PubKey) error {
```

bool is useless here, it's always `err != nil`

go/pkg/bertyprotocol/store_metadata.go
```diff
@@ -330,13 +355,29 @@ func (m *metadataStore) GetIncomingContactRequestsStatus() (bool, *bertytypes.Sh
...
+
+	if m.typeChecker(isContactGroup) {
+		return m.Index().(*metadataStoreIndex).MemberCount()
+	}
+
+	if m.typeChecker(isMultiMemberGroup) {
+		return m.Index().(*metadataStoreIndex).MemberCount()
+	}
+
+	return 0
```
moul, line 371:
```suggestion
	switch {
	case m.typeChecker(isAccountGroup):
		return 0
	case m.typeChecker(isContactGroup):
		return m.Index().(*metadataStoreIndex).MemberCount()
	case m.typeChecker(isMultiMemberGroup):
		return m.Index().(*metadataStoreIndex).MemberCount()
	default:
		// FIXME: log "MemberCount: unsupported type")
		return 0
	}
```

go/pkg/bertyprotocol/group_context.go
```diff
@@ -14,12 +14,12 @@ type groupContext struct {
 	group           *bertytypes.Group
 	metadataStore   *metadataStore
 	messageStore    *messageStore
-	messageKeystore *MessageKeystore
```
moul, line 17: the next time, I suggest you to make a quick PR just for this instead of keeping this small change in a long-running PR

go/pkg/bertyprotocol/scenario_test.go
```diff
@@ -145,6 +145,9 @@ func TestScenario_MessageContactGroup(t *testing.T) {
 		// Add accounts as contacts
 		addAsContact(ctx, t, tps, tps)
 
+		// Waiting that secrets are sent
+		WaitingForContactReady(ctx, t, tps, tps)
```
moul, line 149:
```suggestion
		waitForContactReady(ctx, t, tps, tps)
```

go/pkg/bertyprotocol/scenario_test.go
```diff
@@ -779,6 +782,64 @@ func addAsContact(ctx context.Context, t *testing.T, senders, receivers []*Testi
 	logTree(t, "duration: %s", 0, false, time.Since(start))
 }
 
+func WaitingForContactReady(ctx context.Context, t *testing.T, senders, receivers []*TestingProtocol) {
+	logTree(t, "Waiting that Senders send secrets to Receivers", 0, true)
+	start := time.Now()
+	var waitingSecretsDuration time.Duration
+
+	for _, sender := range senders {
+		for _, receiver := range receivers {
```
moul, line 791: not sure if this is a slow or not, but if yes, maybe you can use https://pkg.go.dev/golang.org/x/sync/errgroup

go/pkg/bertyprotocol/store_metadata_index.go
```diff
@@ -167,12 +168,39 @@ func (m *metadataStoreIndex) handleGroupAddDeviceSecret(event proto.Message) err
 	}
 
 	if m.ownMemberDevice.device.Equals(senderPK) {
-		m.sentSecrets[string(e.DestMemberPK)] = struct{}{}
+		m.sentSecrets[string(e.DevicePK)+string(e.DestMemberPK)] = struct{}{}
```
moul, line 171: not sure if there is already another mutex in a top-level, but this part may be problematic if run concurrently
in general, when managing maps in async environments, always use mutex around the map manipulation

go/pkg/bertyprotocol/store_metadata_index.go
```diff
@@ -167,12 +168,39 @@ func (m *metadataStoreIndex) handleGroupAddDeviceSecret(event proto.Message) err
...
+	ownDeviceRaw, err := m.ownMemberDevice.device.Raw()
+	if err != nil {
+		return nil, errcode.ErrDeserialization.Wrap(err)
+	}
+
+	m.logger.Debug("watchSentSecrets", zap.Any("ownDevice", ownDeviceRaw), zap.Any("memberPK", memberRaw))
+
+	channel, ok := m.chanSentSecrets[string(ownDeviceRaw)+string(memberRaw)]
+	if !ok {
+		channel = make(chan struct{}, 1)
```
moul, line 197:
```suggestion
		channel = make(chan struct{}, 0)
```

to be tested, but would be better to have an unbuffered channel if possible

go/pkg/bertyprotocol/store_metadata_index.go
```diff
@@ -167,12 +168,39 @@ func (m *metadataStoreIndex) handleGroupAddDeviceSecret(event proto.Message) err
...
+		return nil, errcode.ErrDeserialization.Wrap(err)
+	}
+	ownDeviceRaw, err := m.ownMemberDevice.device.Raw()
+	if err != nil {
+		return nil, errcode.ErrDeserialization.Wrap(err)
+	}
+
+	m.logger.Debug("watchSentSecrets", zap.Any("ownDevice", ownDeviceRaw), zap.Any("memberPK", memberRaw))
+
+	channel, ok := m.chanSentSecrets[string(ownDeviceRaw)+string(memberRaw)]
```
moul, line 195:
```suggestion
 	mappingKey := fmt.Sprintf("%s%s", ownDeviceRaw, memberRaw)
	channel, ok := m.chanSentSecrets[mappingKey]
```

reusable few lines after

go/pkg/bertyprotocol/store_message.go
```diff
@@ -40,7 +44,80 @@ func (m *messageStore) setLogger(l *zap.Logger) {
 	m.logger = l.With(zap.String("group-id", fmt.Sprintf("%.6s", base64.StdEncoding.EncodeToString(m.g.PublicKey))))
 }
 
-func (m *messageStore) openMessage(ctx context.Context, e ipfslog.Entry) (*bertytypes.GroupMessageEvent, error) {
+func (m *messageStore) queueScanCache(devicePK []byte) bool {
```
moul, line 47: can you try finding a more explicit name?

go/pkg/bertyprotocol/store_metadata_index.go
D4ryl00: In my case, I know the chan will be used only for one message. The buffer will prevent to block `handleGroupAddDeviceSecret` in `store_metadata_index.go` if the channel is not being read.

go/pkg/bertyprotocol/store_metadata_index.go
D4ryl00: yes handled functions are already protected by a mutex

api/bertytypes.proto
```diff
@@ -565,6 +565,17 @@ message ContactUnblock {
   message Reply {}
 }
 
+message WaitMemberIsReady {
+  message Request {
+    // group_pk is the identifier of the group
+    bytes group_pk = 1 [(gogoproto.customname) = "GroupPK"];
+    // device_pk is the member's device
```
phanhuynhquy, line 572: typo error?

go/pkg/bertyprotocol/store_message.go
```diff
@@ -40,7 +44,80 @@ func (m *messageStore) setLogger(l *zap.Logger) {
...
+			evt, err := m.openMessage(ctx, e, false)
+			if err != nil {
+				m.logger.Error("unable to open message", zap.Error(err))
+				continue
+			}
+
+			m.Emit(ctx, evt)
+			m.unsafeDeleteFromCache(devicePK, e)
+		}
+		if len(events) == 0 {
```
phanhuynhquy, line 100: this block of if statement is redudant?

go/pkg/bertyprotocol/store_message.go
```diff
@@ -147,10 +228,12 @@ func constructorFactoryGroupMessage(s *BertyOrbitDB) iface.StoreConstructor {
...
-			devKS:  s.deviceKeystore,
-			mks:    s.messageKeystore,
-			g:      g,
-			logger: s.Logger(),
+			devKS:     s.deviceKeystore,
+			mks:       s.messageKeystore,
+			g:         g,
+			logger:    s.Logger(),
+			cache:     make(map[string][]ipfslog.Entry),
+			cacheChan: make(chan []byte, 10),
```
phanhuynhquy, line 236: nit: 10 is so small? how to determine this?

go/pkg/bertyprotocol/store_message.go
D4ryl00: oh yes, there was a for loop here, and I forget to remove this block !

go/pkg/bertyprotocol/store_message.go
D4ryl00: This buffer is filled when:
1) a new chainkey is registered
2) a new message is put in the cache while the sender's chainkey is not registered yet

In a real context, these two events are rare, so I decided to put 10.

api/bertytypes.proto
D4ryl00: yes, it's `member_pk`, thx

go/pkg/bertyprotocol/api_event.go
```diff
@@ -176,3 +177,36 @@ func (s *service) GroupMessageList(req *bertytypes.GroupMessageList_Request, sub
...
+	id, err := groupPK.Raw()
+	if err != nil {
+		return errcode.ErrSerialization.Wrap(err)
+	}
+	cg, err := s.getContextGroupForID(id)
+	if err != nil {
+		return errcode.ErrGroupMissing.Wrap(err)
+	}
+
+	err = cg.MetadataStore().WaitSecretsAreSent(sub.Context(), memberPK)
```
phanhuynhquy, line 200: WaitSecretsAreSent() is blocking until all secrets are sent to members. if error happen, it will be stopping until sub's context is cancel. But WaitMemberIsReady() normally is called outside via grpc. I'm not sure we have timeout for gprc handler or not. Do you think in request should have a timeout value to control waiting duration?
for examples, after a timeout, we can try to do a sent missed secrets?

go/pkg/bertyprotocol/store_metadata_index.go
moul: > In my case, I know the chan will be used only for one message. The buffer will prevent to block handleGroupAddDeviceSecret in store_metadata_index.go if the channel is not being read.

even if this is completely possible in some case, by default, **"I know that it uses only one message"** and **"the channel maybe not read"** in the same phrase seems too much fragile

Can you double-check and confirm that you really need a buffer, and if so, that a buffer of 1 is a safe value?

go/pkg/bertyprotocol/store_metadata_index.go
moul: you need to be sure that 100% of the codebase manipulating the map use mutex when manipulating the map

go/pkg/bertyprotocol/api_event.go
phanhuynhquy: `if error happen, it will be stopping until sub's context is cancel`
means that if any abnormal happen, like secret was not sent. so WaitSecretsAreSent() will block forever on waiting data from channel `chanReady`.
And I don't see any control over `sub.ctx` - so this function can block forever.
I'm also not quite familiar with GRPC.
@moul : do you have any idea here?

D4ryl00: The client can call the grpc method with a context.WithTimeout here to cancel : https://github.com/berty/berty/blob/bdd1babf29e06548309e6a125672b17e80bb50c4/go/pkg/bertyprotocol/scenario_test.go#L816

phanhuynhquy: This context is from client side of GRPC, and WaitMemberIsReady() will be called in server side. I'm not sure ctx of client side impact to server side. :).

moul: not sure about the whole discussion, but a thing that is sure, is that we can completely setup a timeout on gRPC, but it may have strange behavior, I prefer each call to have it's own limit so it can decide by itself what is a normal limit and what is not

go/pkg/bertyprotocol/store_message.go
moul: is this limit OK for a conversation of >100 people?

go/pkg/bertyprotocol/api_event.go
phanhuynhquy: but anw, I think it's fine now. Thanks

phanhuynhquy: @moul:
In client side we call WaitMemberIsReady() via GRPC. and in server side, WaitMemberIsReady() will wait forever on `readyChannel` now or until context of handler is done - so we not sure the context lifetime of client side has any impact to context created in handler of server side. That is context I'm discussing.

api/bertyprotocol.proto
```diff
@@ -42,6 +42,9 @@ service ProtocolService {
   // ContactUnblock unblocks a contact from sending requests
   rpc ContactUnblock (types.v1.ContactUnblock.Request) returns (types.v1.ContactUnblock.Reply);
 
+  // WaitMemberIsReady returns a stream that will send a event as soon as the member is ready to receive messages.
+  rpc WaitMemberIsReady (types.v1.WaitMemberIsReady.Request) returns (stream types.v1.WaitMemberIsReady.Reply);
```
aeddi, line 46: This is a bad idea for several reasons:
- It's counter-intuitive and complicates the task for the protocol user, if something like this were to be implemented (waiting for secrets to be sent before sending messages), it would have to be internal to the protocol (using a cache or something), especially since it would have to be systematic.
- It's useless in the sense that the member must be able to receive secrets and messages out of order and have the ability to replay the messages once the secret is received.
- It's also useless as long as the member doesn't send an ACK, the only thing that seems to be done in this call is to check if the "sending" was done in orbit-db. The member has no guarantee to receive the messages before the secret despite this function.

api/errcode.proto
```diff
@@ -44,8 +44,9 @@ enum ErrCode {
   ErrCryptoSignature = 203;
   ErrCryptoSignatureVerification = 204;
   ErrCryptoDecrypt = 205;
-  ErrCryptoEncrypt = 206;
-  ErrCryptoKeyConversion = 207;
+  ErrCryptoDecryptPayload = 206;
```
aeddi, line 47: Why adding `ErrCryptoDecryptPayload` when `ErrCryptoDecrypt` is already available?

go/pkg/bertyprotocol/store_message.go
```diff
@@ -26,10 +27,13 @@ const groupMessageStoreType = "berty_group_messages"
...
 
-	devKS  DeviceKeystore
-	mks    *MessageKeystore
-	g      *bertytypes.Group
-	logger *zap.Logger
+	devKS     DeviceKeystore
+	mks       *messageKeystore
+	g         *bertytypes.Group
+	logger    *zap.Logger
+	cache     map[string][]ipfslog.Entry
```
aeddi, line 34: I think it would be wiser to store the messages in a group scoped circular cache with a fixed size to protect against attack where someone could fill the RAM of all the group members' devices by sending a lot of messages without ever sending the secret.

aeddi: It should not be forgotten that by closing/reopening the group, the problem of messages arrived before the secrets disappears since the replay is done naturally.
We can therefore say that a circular cache of 1024 (for example) messages is sufficient in the sense that if in a highly unlikely case it happens that 1024 messages are received before the corresponding secrets, the problem will be minimal for the user (open/close the group -> problem solved).

aeddi (changes requested):

api/bertyprotocol.proto
D4ryl00: I'll answer your three points individually:

- yes, the only solution is to set a cache for messages before sending them if secrets are not sent.
- I don't agree because replay will not work, because messages sent before secrets sents can never be opened because the sender will send the chainkey already derived.
- That's why there is a message cache for the receiver.

api/errcode.proto
D4ryl00: It's to differentiate why the message can't be opened (`openEnvelopeHeader` ≠ `openPayload` in `keystore_message.go`) before adding it in the cache (I want add it only if `openPayload` failed).

go/pkg/bertyprotocol/store_message.go
D4ryl00: :+1: for the circular cache.

api/bertyprotocol.proto
aeddi: Conclusion after an audio call with D4ryll00:
- This API method in fact does not wait until the member is ready / has received the secrets, it ensures that messages are not sent until the secrets have been sent to prevent the chain key from being derived by sending messages before the chain key is sent as a secret (messages sent before will therefore be undecryptable).
- This API method will be removed and replaced by an internal lock.

go/pkg/bertyprotocol/api_event.go
aeddi: Resolved because this code will be removed, see https://github.com/berty/berty/pull/2312#discussion_r501000681

go/pkg/bertyprotocol/store_metadata_index.go
aeddi: Resolved because this code will be removed, see #2312 (comment)

go/pkg/bertyprotocol/store_metadata_index.go
aeddi: Resolved because this code will be removed, see #2312 (comment)

go/pkg/bertyprotocol/store_metadata.go
aeddi: Resolved because this code will be removed, see #2312 (comment)

go/pkg/bertyprotocol/scenario_test.go
aeddi: Resolved because this code will be removed, see #2312 (comment)

go/pkg/bertyprotocol/group_context.go
aeddi: Done -> https://github.com/berty/berty/pull/2467

go/pkg/bertyprotocol/store_metadata_index.go
```diff
@@ -167,7 +167,7 @@ func (m *metadataStoreIndex) handleGroupAddDeviceSecret(event proto.Message) err
 	}
 
 	if m.ownMemberDevice.device.Equals(senderPK) {
-		m.sentSecrets[string(e.DestMemberPK)] = struct{}{}
+		m.sentSecrets[string(e.DevicePK)+string(e.DestMemberPK)] = struct{}{}
```
aeddi, line 170: Why are you making this change?
This map only concerns our current device and is only modified after checking that the device key of the sender matches.

go/pkg/bertyprotocol/store_metadata_index.go
```diff
@@ -289,7 +289,12 @@ func (m *metadataStoreIndex) areSecretsAlreadySent(pk crypto.PubKey) (bool, erro
 		return false, errcode.ErrInvalidInput.Wrap(err)
 	}
 
-	_, ok := m.sentSecrets[string(key)]
+	ownPK, err := m.ownMemberDevice.device.Raw()
+	if err != nil {
+		return false, errcode.ErrSerialization.Wrap(err)
+	}
+
+	_, ok := m.sentSecrets[string(ownPK)+string(key)]
```
aeddi, line 297: Same thing here: why add our device key to the key of a map that only concerns our current device?

PR #2313, opened 2020-09-14
feat: Allows the selection of key's params using rdvp.

Jorropo:
Also changed the default to `Ed25519` not `RSA` (should be cheaper to compute for mobile).
I think it's worth to change the main rdvp server to `Ed25519` or `ECDSA` now (while not many people are testing berty) (because that change the peer ID).


go/cmd/rdvp/main.go
```diff
@@ -89,7 +94,8 @@ func main() {
 					return errcode.TODO.Wrap(err)
 				}
 			} else {
-				priv, _, err = libp2p_ci.GenerateKeyPairWithReader(libp2p_ci.RSA, 2048, crand.Reader) // nolint:staticcheck
+				// Don't use key params here, this is a dev tool, a real instalations should use a static key.
```
moul, line 97: s/instalations/installation

moul (approved):
And yes, @gfanton, can you update the current dev server's key?

go/cmd/rdvp/main.go
Jorropo: @moul done :)

moul (approved):

go/cmd/rdvp/main.go
```diff
@@ -89,7 +94,8 @@ func main() {
 					return errcode.TODO.Wrap(err)
 				}
 			} else {
-				priv, _, err = libp2p_ci.GenerateKeyPairWithReader(libp2p_ci.RSA, 2048, crand.Reader) // nolint:staticcheck
+				// Don't use key params here, this is a dev tool, a real installation should use a static key.
+				priv, _, err = libp2p_ci.GenerateKeyPairWithReader(libp2p_ci.Ed25519, 2048, crand.Reader) // nolint:staticcheck
```
moul, line 98:
```suggestion
				priv, _, err = libp2p_ci.GenerateKeyPairWithReader(libp2p_ci.Ed25519, -1, crand.Reader) // nolint:staticcheck
```

less ambiguous

moul (approved):

PR #2316, opened 2020-09-15
feat: support multiple discovery drivers at the same time

Jorropo:


Jorropo:
Sucessfuly tested with 2 iphones

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +86,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
+	if err != nil {
+		return nil, err
+	}
+	limit := options.Limit
+	if limit == 0 || limit > maxLimit {
+		limit = maxLimit
+	}
 
+	// Buffer to try to limit context switch while returning peers.
+	outPeers := make(chan p2p_peer.AddrInfo, 10)
```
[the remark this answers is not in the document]
Jorropo: It's a round number. I've just avoided `-1` because I remember it as computationally intensive (might have changed since then) and `0` because that generates lots of contexts switch.
The buffer could be removed.

go/internal/tinder/driver_rdv.go
```diff
@@ -171,6 +171,107 @@ func (c *rendezvousDiscovery) FindPeers(ctx context.Context, ns string, opts ...
...
+				newCacheSize--
+				delete(cache.recs, p)
+			}
+		}
+
+		cookie := cache.cookie
+
+		// Discover new records if we don't have enough
+		if newCacheSize < limit {
+			// TODO: Should we return error even if we have valid cached results?
```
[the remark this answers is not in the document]
Jorropo: It's not by me btw, it just got caught in the copy paste :
https://github.com/berty/berty/blob/b489e357b902b0a1b3bb2449b45ffa2768fd8ec6/go/internal/tinder/driver_rdv.go#L124
I know that redundant code but there is no clean way to merge them without introducing extra of unintuitive branching (it could have a `findPeers` method with an `sync bool` params, `FindPeers` would run it syncronously with `sync == true` and reverse for `FindPeersAsync`, and the sync params would be used to trigger or not some things (goroutine or regular call of the lambada, closing of the chans or not and silent death vs loud one, ...)

go/internal/tinder/driver_multi_test.go
```diff
@@ -135,6 +136,81 @@ func TestMultiDriver_FindPeers(t *testing.T) {
...
+		},
+	}
+
+	logger, cleanup := testutil.Logger(t)
+	defer cleanup()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	ms := NewMockedDriverServer()
+	mn := p2p_mock.New(ctx)
```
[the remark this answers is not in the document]
Jorropo: I wasn't able to as cleanly as I wish because when using `t.Parallel` defered calls are not executed, but what ever it works.

go/internal/tinder/driver_mocked.go
```diff
@@ -82,6 +82,46 @@ func (s *MockDriverServer) FindPeers(ns string, limit int) (<-chan p2p_peer.Addr
 	return ch, nil
 }
 
+func (s *MockDriverServer) FindPeersAsync(ctx context.Context, outChan chan<- p2p_peer.AddrInfo, ns string, limit int) {
```
phanhuynhquy, line 85: FindPeers() and FindPeersAsync() have some same code, could you refractor it?

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +84,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
+						seenLock.RUnlock()
+						continue MainSearchLoop
+					}
+				}
+				seenLock.RUnlock()
+				seenLock.Lock()
+				// We need to check if newer peers were added while we were not locking.
+				curlen := len(alreadySeen)
+				if curlen > limit {
+					return
```
phanhuynhquy, line 172: missing seenLock.Unlock() here

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +84,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
+				}
+				for oldlen < curlen {
+					if alreadySeen[oldlen] == id {
```
phanhuynhquy, line 175: don't really know idea here.
But because oldlen < curlen, and maybe alreadySeen[oldlen] != id.
but maybe id will happen in  alreadySeen[oldlen+1] to curlen-1 ?
Why don't use map here with key is id, value is bool? This will make code simpler than?

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +84,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
+	ndrivers := len(md.drivers)
+	// We don't use a lock array because what we execute next is so cheap it's not worth the extra context switches that generates.
+	var wg sync.WaitGroup
+	wg.Add(ndrivers)
+	errReturn := make([]error, ndrivers)
+	durationReturn := make([]time.Duration, ndrivers)
+	for i := 0; i < ndrivers; i++ {
+		go func(j int) {
+			durationReturn[j], errReturn[j] = md.drivers[j].Advertise(ctx, ns, opts...)
+			wg.Done()
```
phanhuynhquy, line 96: nit: prefer to call defer wg.Done() at begining of function.

go/internal/tinder/driver_multi.go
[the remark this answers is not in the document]
phanhuynhquy: I think 10 should be change to `limit`?

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +84,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
+	var alreadySeen []p2p_peer.ID
+	var seenLock sync.RWMutex
+	for i := 0; i < ndrivers; i++ {
+		go func(j int) {
+			defer wg.Done()
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+			var inPeers <-chan p2p_peer.AddrInfo
+			inPeers, errReturn[j] = md.drivers[j].FindPeers(ctx, ns, opts...)
+			errWg.Done()
```
phanhuynhquy, line 155: should check errReturn[j] here to stop further processing if error happen?

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +84,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
-				selCases[idx].Chan = reflect.ValueOf(nil)
-				ndrivers--
-				continue
-			}
-
-			// we can safly get our peer
-			peer := value.Interface().(p2p_peer.AddrInfo)
-			// fmt.Printf("[multi] found a peers: %v for %s\n", peer, ns)
+	// Buffer to try to limit context switch while returning peers.
+	inPeers := make(chan p2p_peer.AddrInfo, 10)
```
phanhuynhquy, line 226: same above, 10 should be `Limit`.
And it seem that the maxLimit is quite big. Should use a defaultLimit (10 for example) in case of limit is out of range?

go/internal/tinder/driver_multi.go
```diff
@@ -48,126 +84,189 @@ func (md *MultiDriver) Advertise(ctx context.Context, ns string, opts ...p2p_dis
...
+						case <-ctx.Done():
+							return
+						}
+					case <-ctx.Done():
+						return
+					}
+					peerCount++
+				}
+				cancelSearch()
+			}()
```
phanhuynhquy, line 264: missing close channel inPeers

go/internal/tinder/driver_multi.go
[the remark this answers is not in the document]
Jorropo: > I think 10 should be change to `limit`?

Any value would work, I was just trying to not be wasteful on memory, so I've tested and I can't see any difference without creating any realistic setup.

go/internal/tinder/driver_multi.go
Jorropo: The goal is to have at least one finder working, in case all fails the context is cancelled.

go/internal/tinder/driver_multi.go
Jorropo: I don't really care, the context is cancelled and all goroutines having a reference to `inPeers` are about to die.
Also that would add an other waitGroup to avoid `panic: send on closed channel`, I prefer to let the GC take care of that.

go/internal/tinder/driver_multi.go
Jorropo: > And it seem that the maxLimit is quite big. Should use a defaultLimit (10 for example) in case of limit is out of range?

I already ceil the limit if it is out of range :
```go
	limit := options.Limit
	if limit == 0 || limit > maxLimit {
		limit = maxLimit
	}
```

go/internal/tinder/driver_mocked.go
Jorropo: I could but I don't find that more readable.
That would require to create a :
`FindPeersAsync(context.Context, chan<- p2p_peer.AddrInfo, string, ...p2p_discovery.Option) (<-chan p2p_peer.AddrInfo, error)` and this method would check the params chan agaisnt `nil` to know if it have to instanciate a chan, close it, ...

moul (approved):

moul (approved):
LGTM 👍 

be sure to be rebased with master, and squash some commits, please 🙏

Jorropo:
@moul done ! That should now be atomic.

go/internal/tinder/driver_multi.go
phanhuynhquy: yes, please discard this review, I mistake inPeers or outPeers.

PR #2323, opened 2020-09-16
feat: Add lifecycle manager to bertymessenger

gfanton:
Add lifecycle to bertymessenger

### DIFF
--- a/go/framework/bertybridge/bridge_protocol.go
+++ b/go/framework/bertybridge/bridge_protocol.go
@@ -11,6 +11,16 @@ import (
 	"sync/atomic"
 	"time"
 
+	"berty.tech/berty/v2/go/internal/config"
+	"berty.tech/berty/v2/go/internal/ipfsutil"
+	"berty.tech/berty/v2/go/internal/lifecycle"
+	mc "berty.tech/berty/v2/go/internal/multipeer-connectivity-transport"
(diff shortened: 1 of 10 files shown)


go/framework/bertybridge/bridge_protocol.go
```diff
@@ -3,6 +3,7 @@ package bertybridge
 import (
 	"context"
 	"fmt"
+	"math/rand"
```
moul, line 6:
```suggestion
	mrand "math/rand"
```

to follow the convention of the rest of the project

go/internal/lifecycle/manager.go
```diff
@@ -0,0 +1,61 @@
...
+func (m *Manager) UpdateState(state State) {
+	m.condState.L.Lock()
+	if m.currentState != state {
+		m.currentState = state
+		m.condState.Broadcast()
+	}
+	m.condState.L.Unlock()
+}
+
+func (m *Manager) WaitForStateChange(ctx context.Context, state State) (State, error) {
```
moul, line 31: can you add a comment or rename the state parameter to understand how it works

my first feeling was that it will wait the state to be the one expected, but from what I can see, if you wait for the current state, it will block until the value change?

moul (approved):

moul (approved):

PR #2336, opened 2020-09-16
Move the android mobile CI to actions.

Jorropo:


.buildkite/pipeline.yml
```diff
@@ -1,55 +0,0 @@
-steps:
```
moul, line 1: keep buildkite for the moment, we need to build on both for some days to check that everything is ok

.github/workflows/android.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  build:
+    name: Build for Android
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mount bazel cache
+        uses: actions/cache@v1
+        with:
+          path: "/home/runner/.cache/bazel"
+          key: bazel
```
moul, line 21: we don't use bazel anymore

.buildkite/pipeline.yml
Jorropo: ok

.github/workflows/android.yml
Jorropo: ah ok, this was on the install script of buildkite :) (I havn't finished yet)

Jorropo:
We could speed up the IOS build using [Carthage](https://github.com/Carthage/Carthage) to cache some xcode stuff but I don't think that worth the effort (that something build framework like fastlane).

moul (approved):

PR #2344, opened 2020-09-17
fix: button delete fake datas deleted real datas

clegirar:
Button delete fake datas in devtools deleted real messages datas (deleted just on the store)


n0izn0iz (changes requested):
I don't understand how what is changed would have removed non fake messages
Also the second argument of pickBy is the filtering function so if you remove it you can remove the enclosing pickBy

n0izn0iz (approved):
causes a small memleak but nothing to worry about unless an user creates millions of fake contacts or convs so lgtm as quickfix

PR #2349, opened 2020-09-18
feat: improve beta bot

clegirar:
For improve betabot we have think about some features :
- [x] Scenario of conv with betabot
- [x] The bot have a pull of messages and it will take a random value for send it
- [x] The bot should auto-accept the invitation for multimember conversation
---
- [x] Add modal (js) for betabot

For the modal the link for request the betabot was hardcoded in `js/packages/components/modals/AddBetabot.js` line `149`


go/cmd/betabot/main.go
```diff
@@ -103,15 +105,71 @@ func handleEvent(ctx context.Context, messengerClient bertymessenger.MessengerSe
 			if err != nil {
 				return err
 			}
+		} else if contact.State == bertymessenger.Contact_Established {
+			// When contact was established, send message and a group invitation
+			time.Sleep(5 * time.Second)
```
Jorropo, line 110: The event loop being monothread
https://github.com/berty/berty/blob/a39a7406523a8c83af81e730a336d3c142b9aa9b/go/cmd/betabot/main.go#L74-L76
You are effectively blocking the whole loop for 5 seconds.

Jorropo (changes requested):

Jorropo (approved):
LGTM (I'm not confident, pls get an other review)

go/cmd/betabot/main.go
```diff
@@ -103,15 +101,71 @@ func handleEvent(ctx context.Context, messengerClient bertymessenger.MessengerSe
...
```
moul, line 106: I think you should use time.After here

Jorropo: I don't think, `time.After` is way costier than `Sleep` and this code is running in a goroutine so no need of it.

go/cmd/betabot/main.go
```diff
@@ -103,15 +101,71 @@ func handleEvent(ctx context.Context, messengerClient bertymessenger.MessengerSe
...
+			}
+			_, err = messengerClient.Interact(ctx, &bertymessenger.Interact_Request{
+				Type:                  bertymessenger.AppMessage_TypeUserMessage,
+				Payload:               userMessage,
+				ConversationPublicKey: contact.ConversationPublicKey,
+			})
+			if err != nil {
+				return err
+			}
+			<-time.After(5 * time.Second)
```
Jorropo, line 121: That still sync (a chan is async but interacting with it (`<-`) is sync), what is different with chan is you can use them in selects:
```go
select {
case value:=<-channel:
  // code executed when a value was red from the channel
case channel<-valueToWrite:
  // code executed when a value was written from the channel
case <-time.After(time.Second):
  // code executed if nothing was done after 1s
}
```
When a select is done it will first try to to do a non blocking interact with the channels one by one starting from the top (non blocking means if can't interact (due to channel blocked or buffer less) skip the case), if a channel interaction was sucessfull it will execute the case and exit the select with the implied break statement at the end of select and switch case in go.
Then if all non blocking interaction failed the runtime takes the lead, basicaly your goroutine will likely context switch and will be eventualy be wakeup later on a sucessfull case but you can see this as quite random (if your first case would be executable first and then a second one your goroutine might be resumed later with the second or first case).

Note that if you only want the first part of select (checking all cases one by one) but not the runtime sleepy part (waiting until a case resolve) you can add a default case in your select :
```go
select {
case value:=<-channel:
  // code executed when a value was red from the channel
case channel<-valueToWrite:
  // code executed when a value was written from the channel
default:
  // code executed if all other interaction were blocking (we rarely need to put code there but it might be usefull)
}
// Next code
```

go/cmd/betabot/main.go
```diff
@@ -64,19 +65,13 @@ func betabot() error {
...
 				if err != nil {
 					log.Printf("stream error: %v", err)
 				}
-				handlerMutex.Lock()
-				err = handleEvent(ctx, messengerClient, gme)
-				handlerMutex.Unlock()
-				if err != nil {
-					log.Printf("event handler error: %v", err)
-				}
+				handleEvent(ctx, messengerClient, gme)
```
Jorropo, line 74: still missing something there

go/cmd/betabot/main.go
```diff
@@ -61,22 +63,28 @@ func betabot() error {
...
-				gme, err := s.Recv()
-				if err != nil {
-					log.Printf("stream error: %v", err)
-				}
-				handlerMutex.Lock()
-				err = handleEvent(ctx, messengerClient, gme)
-				handlerMutex.Unlock()
-				if err != nil {
-					log.Printf("event handler error: %v", err)
+				select {
```
Jorropo, line 74: The goal of the select is to wait multiples channel (you should maybe `<-s.Recv()` (doesn't work right now)) :
```go
defer wg.Done()
// Loops and ...
select {
case gme, err := <-s.Recv():
  // handle event
case <-ctx.Done():
  return
}
```

go/cmd/betabot/main.go
```diff
@@ -61,22 +63,28 @@ func betabot() error {
...
+				case <-ctx.Done():
+					wg.Wait()
+				default:
+					gme, err := s.Recv()
+					if err != nil {
+						log.Printf("stream error: %v", err)
+						return
+					}
+					wg.Add(1)
+					defer wg.Done()
```
Jorropo, line 84: you must merge the wg.Done() with the handleEvent :
```go
go func(){
  defer wg.Done()
  handleEvent(ctx, messengerClient, gme)
}()
```

Jorropo (changes requested):

go/cmd/betabot/main.go
```diff
@@ -58,63 +60,136 @@ func betabot() error {
 	}
 
 	// event loop
+	var wg sync.WaitGroup
+	wg.Add(1)
```
Jorropo, line 64: To avoid confusion we usualy try to put the `wg.Add()` right before firing the goroutine (so it's easy to connect the Add and the goroutine) but that just styling, this work just fine.

EDIT: And here that needed, lets assume `messengerClient.EventStream` return an error you are never gonna do a `Done` (because the return stop the current scope never reaching the `go func(){defer wg.Done() // ...`)

Jorropo (changes requested):

Jorropo (approved):
LGTM

go/cmd/betabot/main.go
```diff
@@ -57,64 +61,168 @@ func betabot() error {
...
-				err = handleEvent(ctx, messengerClient, gme)
-				handlerMutex.Unlock()
-				if err != nil {
-					log.Printf("event handler error: %v", err)
-				}
+
+				wg.Add(1)
+				go func() {
+					handleEvent(ctx, messengerClient, gme)
+					wg.Done()
```
Jorropo, line 118: Should be defered before the `handleEvent` (the different is that defered are executed in case of panics)

Jorropo (changes requested):

go/cmd/betabot/main.go
```diff
@@ -57,64 +62,326 @@ func betabot() error {
 		qrterminal.GenerateHalfBlock(res.HTMLURL, qrterminal.L, os.Stdout)
 	}
 
+	var wg sync.WaitGroup
+	// request contact in flag
+	{
+		if *contact != "" {
+			splits := strings.Split(*contact, ",")
+			wg.Add(1)
+			go func() {
```
moul, line 71: no need for async here, the protocol will be async, and by removing the go routine, you can simplify this part of the code by removing the cancel() and by handling the errors directly as initialization errors

go/cmd/betabot/main.go
```diff
@@ -57,64 +62,326 @@ func betabot() error {
...
+						if err != nil {
+							cancel()
+							return
+						}
+						_, err = messengerClient.SendContactRequest(ctx, &bertymessenger.SendContactRequest_Request{BertyID: parseRet.BertyID, OwnMetadata: om})
+						if err != nil {
+							cancel()
+							return
+						}
+					}
```
moul, line 90: just a note, nothing to do in this PR, but I think we can add an helper in the bertymessenger package for this

go/cmd/betabot/main.go
```diff
@@ -57,64 +62,326 @@ func betabot() error {
...
+			}()
+		}
+	}
+
 	// event loop
 	{
 		s, err := messengerClient.EventStream(ctx, &bertymessenger.EventStream_Request{})
 		if err != nil {
+			log.Printf("Failed to listen EventStream %v", err)
 			return err
```
moul, line 101:
```suggestion
			return fmt.Errorf("failed to listen to EventStream: %w", err)
```

go/cmd/betabot/main.go
```diff
@@ -57,64 +62,326 @@ func betabot() error {
...
+				"OK, copy that.",
+				"OK, I understand.",
+				"Hmmm, Hmmmm. One more time ?",
+				"I think you're the most clever human I know.",
+				"I missed you babe.",
+				"OK, don't send me nudes, I'm a bot dude.",
+				"Come on, let's party.",
+				"May we have a chat about our love relationship future ?",
+				"That's cool. I copy.",
+			} // 28
```
moul, line 351: to improve readability, move this in a dedicated function at the bottom of the file

go/cmd/betabot/main.go
```diff
@@ -57,64 +62,326 @@ func betabot() error {
...
+					return
+				}
+				return
+			}
+			// auto-reply to user's messages
+			nBig, err := rand.Int(rand.Reader, big.NewInt(28))
+			if err != nil {
+				log.Printf("Failed to generate randome number: %v", err)
+				return
+			}
```
moul, line 378: I suggest you to init the seed with `rand.Seed(srand.MustSecure())` at the top of the betabot() function

then you can just use `answers[rand.Intn(len(answers))]`

go/cmd/betabot/main.go
```diff
@@ -123,20 +390,44 @@ func handleEvent(ctx context.Context, messengerClient bertymessenger.MessengerSe
 				ConversationPublicKey: interaction.ConversationPublicKey,
 			})
 			if err != nil {
-				return err
+				log.Printf("handle event: %v", err)
+				return
+			}
+		} else if interaction.Type == bertymessenger.AppMessage_TypeGroupInvitation && !interaction.IsMe {
```
moul, line 396: replace `if`, `else if`, with `switch` statement

go/cmd/betabot/main.go
Jorropo: I guess an global object would be better and simpler.

go/cmd/betabot/main.go
```diff
@@ -123,20 +390,44 @@ func handleEvent(ctx context.Context, messengerClient bertymessenger.MessengerSe
...
-
 	default:
 		log.Printf("<<< %s: ignored", gme.Event.Type)
 	}
-	return err
 }
 
-func waitForCtrlC() {
+func checkValidationMessage(s string) bool {
+	return s == "Yes" || s == "yes" || s == "y" || s == "Y" || s == "YES"
```
moul, line 419:
```suggestion
	switch strings.Lower(s) {
		case "y", "yes", "yes!":
			return true
		default:
			return false
	}
```

Jorropo, line 419:
```suggestion
	ls := strings.ToLower(s)
	return ls == "yes" || ls == "y"
```

go/cmd/betabot/main.go
```diff
@@ -57,64 +62,326 @@ func betabot() error {
...
-				err = handleEvent(ctx, messengerClient, gme)
-				handlerMutex.Unlock()
-				if err != nil {
-					log.Printf("event handler error: %v", err)
-				}
+
+				wg.Add(1)
+				go func() {
+					handleEvent(ctx, messengerClient, gme)
+					wg.Done()
```
Jorropo, line 118:
```suggestion
					defer wg.Done()
					handleEvent(ctx, messengerClient, gme)
```
Safer in case of panic (defer are executed when a panic occurs)

Jorropo (changes requested):

go/cmd/betabot/main.go
```diff
@@ -57,64 +63,289 @@ func betabot() error {
...
+
+			answers := getAnswersRange()
+			// auto-reply to user's messages
+			if err != nil {
+				log.Printf("Failed to generate randome number: %v", err)
+				return
+			}
 			userMessage, err := proto.Marshal(&bertymessenger.AppMessage_UserMessage{
-				Body: "Welcome to the beta!",
+				Body: answers[rand.Intn(len(answers))], // nolint:gosec // we need to use math/rand here, but it is seeded from crypto/rand
```
Jorropo, line 344: It's really an not important random action so using an unseeded `rand` would be fine.
What gosec wants you to do is not seeding it correctly (even if that a good thing) but he wants you to use something like `math/rand#Rand` (obtained with `math/rand#New`), this isolate this random loop from others (avoiding correlation attacks) and allows for concurrent random generation (the way computers generate randomness is sequential, each thread having his own random loop allows for concurrency).

go/cmd/betabot/main.go
```diff
@@ -123,20 +354,49 @@ func handleEvent(ctx context.Context, messengerClient bertymessenger.MessengerSe
...
 }
 
-func waitForCtrlC() {
+func checkValidationMessage(s string) bool {
+	switch strings.ToLower(s) {
+	case "y", "yes", "yes!":
+		return true
+	default:
+		return false
+	}
```
Jorropo, line 388:
```suggestion
	}
	return false
```

Jorropo (approved):
SGTM

PR #2354, opened 2020-09-20
fix: improve leakage cleanup in manager_test.go

phanhuynhquy:
### [Change description]
Improve clean up in manager.go

### [Test]
Run test sets again


go/internal/initutil/manager.go
```diff
@@ -143,18 +145,37 @@ func (m *Manager) Close() error {
...
 
+	if m.Node.GRPC.bufServer != nil {
+		m.Node.GRPC.bufServer.Stop()
+	}
+
 	if m.Node.GRPC.bufServerListener != nil {
 		m.Node.GRPC.bufServerListener.Close()
 	}
+
+	if m.Node.GRPC.clientConn != nil {
```
moul, line 156: Shouldn't this be called before closing the bufserver ?

go/internal/initutil/manager_test.go
```diff
@@ -8,15 +8,54 @@ import (
 	"testing"
 	"time"
 
+	"berty.tech/berty/v2/go/internal/initutil"
```
moul, line 11: That's strange that the CI didn't complain, because I configured golangci-lint to use "gci";

Can you check that you have rebased against master recently and run "make go.fmt" from the go dir?

moul (changes requested):

moul (approved):
LGTM

go/internal/initutil/manager_test.go
```diff
@@ -163,20 +202,14 @@ func TestCloseByContext(t *testing.T) {
 	server, err := manager.GetLocalProtocolServer()
 	require.NoError(t, err)
 	require.NotNil(t, server)
+
+	cancel()
 }
 
 func TestUnstableFlagsLeak(t *testing.T) {
```
moul, line 209:
```suggestion
func TestFlagsLeak(t *testing.T) {
```

go/internal/initutil/manager_test.go
```diff
@@ -146,10 +186,9 @@ func TestTwoConcurrentManagers(t *testing.T) {
 }
 
 func TestCloseByContext(t *testing.T) {
-	// FIXME: defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
-
+	defer verifyRunningLeakDetection(t)
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
```
moul, line 152: I don't get why you move this defer call at the end without defer?

If I'm not wrong, this defer will be stacked and be called "before" the "verifyRunning" one, so for me, it's the same behavior, except that cancel will be called even if a "require.something" make the call return earlier

phanhuynhquy: Yes, at here I did not sure the order of deferred functions are called.
So that why move cancel() to end without defer to ensure cancel() is called before verifyRunningLeakDetection() is called.

phanhuynhquy: Reverted it.

PR #2363, opened 2020-09-22
fix: some UI/UX stuff

clegirar:
- [x] Set home screen as initial screen when app loaded
- [x] Fix responsiveness of background image sample when we haven't any requests or convs
- [x] Fix top padding in 1to1 conv
---
- [x] Expand clickable zone of the arrow for `go back` in some screens
- [x] Expand clickable zone of button `send message` in conv
---
- [x] Fix buck in betabot modal


zxxma:
LGTM 
+ fixed Hex 383B62  on main text police.

PR #2364, opened 2020-09-23
chore: Updated IPFS to v0.7.0

Jorropo:
`go.mod` temporary replace can be removed once ipld/go-car#34 is merged and released.


moul (changes requested):
missing a `make depaware.generate`

Jorropo:
> missing a `make depaware.generate`
@moul 

Fixed :) (I'll add a depaware check to the CI if that not too time costy)

moul:
> Fixed :) (I'll add a depaware check to the CI if that not too time costy)

no, please don't, it will make the development process harder for not so much

I just want the command to be run each time we make massive dependency bumps

Jorropo:
> no, please don't
@moul 

too late :D yes I see, I havn't thinked of that.

moul (approved):

PR #2365, opened 2020-09-24
fix: replication server token authentication

gfanton:
fix token auth on replication server

Signed-off-by: Guilhem Fanton <guilhem.fanton@gmail.com>


glouvigny (approved):

go/internal/initutil/node.go
```diff
@@ -41,6 +45,12 @@ func (m *Manager) SetupLocalProtocolServerFlags(fs *flag.FlagSet) {
 	// p2p.no-ble
 }
 
+func (m *Manager) SetupProtocolAuth(fs *flag.FlagSet) {
+	m.Node.Protocol.needAuth = true
```
moul, line 49: I think that also made the same mistake, but this configures the manager globally, so that's not the good way to do it, the good way is set this to true in the handler itself (in the main.go), after the parsing, this way, it's only true if you enter in the command

I will try to make "invisible flags" to make this automatically, for now you need to set this manually from the main, sorry

go/internal/initutil/node.go
```diff
@@ -41,6 +45,12 @@ func (m *Manager) SetupLocalProtocolServerFlags(fs *flag.FlagSet) {
...
+	fs.StringVar(&m.Node.Protocol.AuthSecret, "auth.secret", "", "API Authentication Secret (base64 encoded)")
```
moul, line 50: flag string and comment are not clear about "what" you are talking about?

which API?

go/internal/initutil/node.go
```diff
@@ -328,18 +338,33 @@ func (m *Manager) getGRPCServer() (*grpc.Server, *grpcgw.ServeMux, error) {
 		grpc_zap.ReplaceGrpcLoggerV2(grpcLogger)
 		grpcLoggerConfigured = true
 	}
+
+	// noop auth func
+	authFunc := func(ctx context.Context) (context.Context, error) { return ctx, nil }
+
+	if m.Node.Protocol.needAuth {
```
moul, line 345: maybe you can just skip this boolean check to check that secret and pk are not empty? this will avoid updaitng the main.go file to set "needAuth" to true

moul (approved):

PR #2367, opened 2020-09-24
wip: fix: messenger proto improvements

glouvigny:


api/bertymessenger.proto
```diff
@@ -264,18 +264,18 @@ message Account {
 }
 
 message Interaction {
-  string cid = 1 [(gogoproto.moretags) = "gorm:\"uniqueIndex\"", (gogoproto.customname) = "CID"];
-  AppMessage.Type type = 2;
+  string cid = 1 [(gogoproto.moretags) = "gorm:\"uniqueIndex\"", (gogoproto.customname) = "CID"]; // FIXME(?): primaryKey instead of uniqueIndex?
```
n0izn0iz, line 267: If you find a way to have both a primary key and a secondary autoincrement index I would gladly make the change

api/bertymessenger.proto
```diff
@@ -284,7 +284,7 @@ message Contact {
   Conversation conversation = 3;
   State state = 4;
   string display_name = 5;
-  int64 created_date = 7;
+  int64 created_date = 7; // FIXME(?): If this is not related to the row creation date should we unify this field name with gorm.Model (ID, CreatedAt, UpdatedAt, DeletedAt)
```
n0izn0iz, line 287: I wanted to avoid the behavior changes that gorm.Model adds

api/bertymessenger.proto
```diff
@@ -264,18 +264,18 @@ message Account {
...
-  string conversation_public_key = 3;
+  string conversation_public_key = 3; // FIXME(?): add index
   Conversation conversation = 4;
   bytes payload = 5;
   bool is_me = 6;
   int64 sent_date = 9;
-  bool acknowledged = 10;
-  int64 arrival_index = 11 [(gogoproto.moretags) = "gorm:\"autoIncrement\""];
+  bool acknowledged = 10; // FIXME(?): By whom? Would a timestamp be better?
+  int64 arrival_index = 11 [(gogoproto.moretags) = "gorm:\"autoIncrement\""]; // FIXME(?): I think this implies a primary key field, can't we set a timestamp instead?
```
n0izn0iz, line 278: yes but as you pointed system clocks can be off.
this isz only used to get the "acks/messages without device" backlog in the correct order

api/bertymessenger.proto
```diff
@@ -264,18 +264,18 @@ message Account {
...
   Member member = 8;
-  string conversation_public_key = 3;
+  string conversation_public_key = 3; // FIXME(?): add index
   Conversation conversation = 4;
   bytes payload = 5;
   bool is_me = 6;
   int64 sent_date = 9;
-  bool acknowledged = 10;
-  int64 arrival_index = 11 [(gogoproto.moretags) = "gorm:\"autoIncrement\""];
+  bool acknowledged = 10; // FIXME(?): By whom? Would a timestamp be better?
```
n0izn0iz, line 277: I think that if we want more informations on acks we should keep the ack interactions in the db and do a join on the target CID

api/bertymessenger.proto
```diff
@@ -190,7 +190,7 @@ message AppMessage {
     TypeUserReaction = 2;
     TypeGroupInvitation = 3;
     TypeSetGroupName = 4;
-    TypeSetUserName = 5;
+    TypeSetUserName = 5; // FIXME: SetDisplayName ?
```
n0izn0iz, line 193: the group also has a DisplayName, if you really want to add "Display", I suggest:
```
TypeSetGroupDisplayName = 4;
TypeSetUserDisplayName = 4;
```

n0izn0iz: or `TypeSetDisplayName` with a `kind` field instead of the two

api/bertymessenger.proto
```diff
@@ -253,8 +253,8 @@ message ConversationJoin {
 message Account {
   string public_key = 1 [(gogoproto.moretags) = "gorm:\"primaryKey\""];
   string display_name = 2;
-  string link = 3;
-  State state = 4;
+  string link = 3; // FIXME(?): Store RDV seed and generate it instead, provide it in the struct returned via the API
+  State state = 4; // FIXME(?): What is this, how is it used, does it have to be stored in the database
```
n0izn0iz, line 257: This is used to signal to the app that the account has been setup and is ready to be used when in Ready state, currently the code that sets it only verifies that the account has a custom name

api/bertymessenger.proto
```diff
@@ -294,28 +294,28 @@ message Contact {
...
 message Conversation {
-  string public_key = 1 [(gogoproto.moretags) = "gorm:\"primary_key\""];
-  Type type = 2;
-  bool is_open = 3;
+  string public_key = 1 [(gogoproto.moretags) = "gorm:\"primaryKey\""];
+  Type type = 2; // FIXME(?): add index
+  bool is_open = 3; // FIXME(?): remove from database
   string display_name = 4;
-  string link = 5;
+  string link = 5; // FIXME(?): avoid persisting it in the database
```
n0izn0iz, line 306: why ?

api/bertymessenger.proto
```diff
@@ -294,28 +294,28 @@ message Contact {
...
-  string contact_public_key = 8;
+  string contact_public_key = 8; // FIXME(?): Add index
   // specific to ContactType conversations
   Contact contact = 9;
   // specific to MultiMemberType conversations
   repeated Member members = 10;
-  string account_member_public_key = 11;
-  string local_device_public_key = 12;
-  int64 created_date = 13;
+  string account_member_public_key = 11; // FIXME: What is the purpose of this field?
```
n0izn0iz, line 316: this is the memberPK of the current account for this conversation

api/bertymessenger.proto
```diff
@@ -294,28 +294,28 @@ message Contact {
...
+  string local_device_public_key = 12; // FIXME: What is the purpose of this field?
```
n0izn0iz, line 317: this is the devicePK of the current device for this conversation

PR #2375, opened 2020-09-25
chore: add betabot store + various go style improvements

moul:
@clegirar, can you check that things are working as before?

I think yes, but I didn't make a full series of tests

---

@gfanton I updated the docker-compose.yml file, but I think it will conflict with some uncommited configuration


clegirar (approved):
LGTM 👍

PR #2381, opened 2020-09-28
feat: Allow swiping back on the MyBertyId view.

Jorropo:
I'm testing that right now but I think I'll have to use a dispatch for this to works correctly.

### DIFF
--- a/js/packages/components/settings/MyBertyId.tsx
+++ b/js/packages/components/settings/MyBertyId.tsx
@@ -2,13 +2,15 @@ import React, { useState } from 'react'
 import { View, TouchableOpacity, ScrollView, Share } from 'react-native'
 import { Layout, Text, Icon } from 'react-native-ui-kitten'
 import { useStyles } from '@berty-tech/styles'
+import { CommonActions } from '@react-navigation/native'
 import { TabBar } from '../shared-components/TabBar'
 import { RequestAvatar } from '../shared-components/Request'
 import { Messenger } from '@berty-tech/store/oldhooks'
-import { useNavigation } from '@berty-tech/navigation'
+import { useNavigation, Routes } from '@berty-tech/navigation'
 import QRCode from 'react-native-qrcode-svg'
 import { FingerprintContent } from '../shared-components/FingerprintContent'
 import { SafeAreaView } from 'react-native-safe-area-context'
+import { SwipeNavRecognizer } from '../shared-components/SwipeNavRecognizer'
 
 //
 // Settings My Berty ID Vue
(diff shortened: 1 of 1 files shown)


Jorropo:
It have a bad feeling to use and I can't make it better (I'm bad at JS) @ekelen I know you did the swipe navigation, could you pls take look when you have time ? (feel free to scrap, it's nothing and doesn't work properly).

ekelen:
It works fine on my simulator (though I would make it a downward caret and have the animation vertical). What isn't working about it on your end?

Jorropo:
> It works fine on my simulator (though I would make it a downward caret and have the animation vertical). What isn't working about it on your end?

@ekelen 

- There is no progression when swiping (if you swipe on a correctly made component **while you are swipping** the destination is revealed bellow your current page, here only when the swipe is done you are moved).
- The page goes down instead of left (as you pointed).

ekelen:
> > It works fine on my simulator (though I would make it a downward caret and have the animation vertical). What isn't working about it on your end?
> 
> @ekelen
> 
>     * There is no progression when swiping (if you swipe on a correctly made component **while you are swipping** the destination is revealed bellow your current page, here only when the swipe is done you are moved).

Good catch; on my sim it's too fast for me to notice. I can check it out later today or tomorrow.

>     * The page goes down instead of left (as you pointed).

Happy to fix (same timeline) if you like, should be fine :)

Jorropo:
> Good catch; on my sim it's too fast for me to notice. I can check it out later today or tomorrow.

If you want to test you can consciently hold your finger (or mouse I guess) while swiping.

> Happy to fix (same timeline) if you like, should be fine :)

I would like this a lot (I really can't do better right now), I've set allow edits to maintainers, you should be able to do it :)

ekelen:
I changed my mind about the vertical swipe; we better keep it as a horizontal swipe, unless we make it a true vertical modal (like HomeModal).

Sorry for the video screencap; this is halfway through downward drag.

image

What do you think if I leave it as a horizontal swipe (motion + animation), but expose underlying screen during swipe as discussed?

Jorropo:
> What do you think if I leave it as a horizontal swipe (motion + animation), but expose underlying screen during swipe as discussed?

@ekelen
Depends :
- If you need to start your scroll at the top edge (like for left and right), I don't find this very intuitive (I find this requiring way to many finger movement) but it shouldn't feel broken anymore.
I guess to be more intuitive this could be made consistant with other vertical popping screen (right now I think of `Scan QR code`, `New Group`, ...)
- If you can swipe from any dead content (text, image, bg color, ...) like on almost anything on Android it should be fine (but I guess that also applies to left and right swiping plus I think it's more my personal habits of only touching the bottom tier of the screen talking).

ekelen (commented):
Looks good, I'm adding background previous nav stack screen as discussed (and vertical direction after all)

Jorropo:
Closed in #2396

PR #2387, opened 2020-09-28
WIP: fix(js): margin flash above chat text input

ekelen:


ekelen:
@zxxma If you can show me a screenshot/way to recreate this issue, that would be helpful. In conversation with @clegirar I'm not sure I've fixed the right thing.

PR #2400, opened 2020-09-30
chore(messenger): db and handlers refactor

glouvigny:


api/bertymessenger.proto
```diff
@@ -326,8 +320,8 @@ message Conversation {
   Contact contact = 9;
   // specific to MultiMemberType conversations
   repeated Member members = 10;
-  string account_member_public_key = 11;
-  string local_device_public_key = 12;
+  string account_member_public_key = 11; // FIXME: Should we keep this?
+  string local_device_public_key = 12; // FIXME: Should we keep this?
```
moul, line 324: not sure this is used, but keeping it makes preloading with gorm super easy

go/pkg/bertymessenger/handlers.go
```diff
@@ -304,536 +439,398 @@ func groupMemberDeviceAdded(svc *service, gme *bertytypes.GroupMetadataEvent) er
...
 
-				// if conv is not open, increment the unread_count
-				if !i.IsMe && !conv.GetIsOpen() {
-					updates["unread_count"] = gorm.Expr("unread_count + 1")
-				}
+func (h *eventHandler) hydrateInteraction(gme *bertytypes.GroupMessageEvent, i *Interaction) error {
+	// fetch conv from db
+	conversation, err := h.db.getConversationByPK(i.ConversationPublicKey)
+	if err != nil {
+		// FIXME: maybe we should accept receiving app messages on unknown conversations?
```
moul, line 737: yes, I think we should keep the warn but not return an error, I'm curious about when it will happen

go/pkg/bertymessenger/handlers.go
```diff
@@ -63,27 +133,26 @@ func accountServiceTokenAdded(svc *service, gme *bertytypes.GroupMetadataEvent)
 		return errcode.ErrDeserialization.Wrap(err)
 	}
 
-	if err := addServiceToken(svc.db, bytesToString(config.AccountPK), ev.ServiceToken); err != nil {
+	if err := h.db.addServiceToken(bytesToString(config.AccountPK), ev.ServiceToken); err != nil {
```
moul, line 136: I know you didn't introduced this `bytesToString` helper, but since this PR is about refactoring, I think we should rename it to something more related to what it actually does

go/pkg/bertymessenger/handlers.go
```diff
@@ -1,59 +1,129 @@
...
+			return err
+		}
+
+		// parse payload
+		amPayload, err := am.UnmarshalPayload()
+		if err != nil {
+			return err
+		}
+
+		i, err := handler.handler(tx, i, amPayload)
```
moul, line 110:
```suggestion
		i, err = handler.handler(tx, i, amPayload)
```

it may be expected, if so, please add a comment explaining why

go/pkg/bertymessenger/db.go
```diff
@@ -48,254 +62,709 @@ func dropAllTables(db *gorm.DB) error {
...
+
+	return d.getDeviceByPK(devicePK)
+}
+
+func (d *dbWrapper) updateContact(contact Contact) error {
+	if contact.PublicKey == "" {
+		return errcode.ErrInvalidInput.Wrap(fmt.Errorf("no public key specified"))
+	}
+
+	return d.tx(func(wrapper *dbWrapper) error {
```
moul, line 545:
```suggestion
	return d.tx(func(tx *dbWrapper) error {
```

This variable is sometimes named `d`, sometimes `db`, sometimes, `wrapper`, what about just nameing it `tx` everywhere?

go/pkg/bertymessenger/db.go
```diff
@@ -48,254 +62,709 @@ func dropAllTables(db *gorm.DB) error {
...
-	return &contact, &conversation, nil
+	return contact, conversation, nil
 }
 
-func addServiceToken(db *gorm.DB, accountPK string, serviceToken *bertytypes.ServiceToken) error {
-	if err := db.Transaction(func(db *gorm.DB) error {
+func (d *dbWrapper) markInteractionAsAcknowledged(cid string) (bool, error) {
+	rowsAffected := int64(0)
+
+	if err := d.tx(func(db *dbWrapper) error {
```
moul, line 436: if I'm not wrong, there is only 1 write, why using a tx here?

go/pkg/bertymessenger/handlers.go
```diff
@@ -1,59 +1,129 @@
...
+	handler, ok := h.appMessageHandlers[i.GetType()]
+
+	if !ok {
+		h.logger.Warn("unsupported app message type", zap.String("type", i.GetType().String()))
+
+		return nil
+	}
+
+	// start a transaction
+	return h.db.tx(func(tx *dbWrapper) error {
```
moul, line 99: ultra nice 👏

moul:
@glouvigny, here is a fix for goreleaser: https://github.com/moul/berty/commit/bc74bcb4700b8069ca440f0852e17791453ce692

moul (approved):
Code LGTM 👍 
Didn't give a real try

go/pkg/bertymessenger/service.go
```diff
@@ -208,17 +207,12 @@ func New(client bertyprotocol.ProtocolServiceClient, opts *Opts) (Service, error
...
 	// subscribe to contact groups
 	{
-		var contacts []Contact
-		err := svc.db.Find(&contacts).Error
+		contacts, err := svc.db.getContactsByState(Contact_OutgoingRequestSent)
 		if err != nil {
 			return nil, err
 		}
 		for _, c := range contacts {
-			if c.State != Contact_OutgoingRequestSent {
```
n0izn0iz, line 217: why remove this ? without this check we will subscribe twice on accepted contacts and subscribe needlessly in other cases

PR #2401, opened 2020-10-01
fix: ignore ipfs api errors

D4ryl00:
Ignore API Ipfs errors to allow two Berty instances in the same place


n0izn0iz (approved):
Code LGTM and tested OK on iOS

PR #2415, opened 2020-10-05
feat: add a package that get username on iOS, Android and others

aeddi:


go/pkg/username/others.go
```diff
@@ -0,0 +1,16 @@
...
+package username
+
+import "os/user"
+
+const defaultUsername = "anon#1337"
+
+func getUsername() string {
+	user, err := user.Current()
+	if err != nil || user == nil {
+		return ""
```
moul, line 12:
```suggestion
		return defaultUsername
```

js/Makefile
```diff
@@ -16,6 +16,7 @@ APP_CONFIG ?= development
 IOS_APP_TARGET ?= Berty Debug
 ANDROID_APP_TARGET ?= Debug
 IOS_DEVICE ?= iPhone 11
+REQUIRED_JAVA_VER ?= 18
```
moul, line 19: shouldn't it be hardcoded instead of configurable?

go/pkg/username/others.go
aeddi: Already handled by the public package function `GetUsername()` in go/pkg/username/username.go:12

js/Makefile
aeddi: Not sure about this, I can hardcode it first and we'll see later?

moul: I'm more interested to have an example of use case where it's a good idea to make this overridable?

moul (approved):

go/pkg/bertymessenger/api.go
```diff
@@ -85,10 +87,9 @@ func (svc *service) internalInstanceShareableBertyID(ctx context.Context, req *I
 		return nil, errcode.TODO.Wrap(err)
 	}
 
-	displayName := req.DisplayName
+	displayName := strings.TrimSpace(req.DisplayName)
 	if displayName == "" {
-		// FIXME: get it from somewhere
-		displayName = "anonymous#1337"
+		return nil, errcode.TODO
```
n0izn0iz, line 92: maybe don't return an error since a client might want to not use a name and the links format supports it

n0izn0iz (approved):
LGTM, we'll do front in another PR

PR #2416, opened 2020-10-05
perf: Cleaner replay check for beta bot and fix some race condition.

Jorropo:
I know perf doesn't count for betabot, but I find this cleaner.

### DIFF
--- a/go/cmd/betabot/main.go
+++ b/go/cmd/betabot/main.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/matryer/resync"
 	qrterminal "github.com/mdp/qrterminal/v3"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
@@ -46,23 +47,52 @@ func main() {
 
 type Bot struct {
 	store struct {
-		Convs       []Conversation
+		Convs       []*Conversation
 		StaffConvPK string
 	}
-	client      bertymessenger.MessengerServiceClient
-	storeIsNew  bool
-	storePath   string
-	storeMutex  sync.Mutex
-	isReplaying bool
-	logger      *zap.Logger
+	client           bertymessenger.MessengerServiceClient
+	storeIsNew       bool
+	storePath        string
+	storeConvMap     map[*Conversation]*sync.Mutex
+	storeConvMapLock sync.Mutex
+	// storeWholeConvLock is a lock where holding it is equivalent to holding all
+	// conversations locks.
+	storeWholeConvLock sync.RWMutex
+	storeMutex         sync.RWMutex
+	storeOnce          resync.Once
+	isReplaying        bool
+	logger             *zap.Logger
 }
 
 type Conversation struct {
 	ConversationPublicKey string
 	ContactPublicKey      string
 	ContactDisplayName    string
-	Count                 int
+	Step                  uint
 	IsOneToOne            bool
+}
+
+// The function returned by LockConversation is the Unlocker.
+func (bot *Bot) LockConversation(c *Conversation) func() {
+	bot.storeConvMapLock.Lock()
+	l, ok := bot.storeConvMap[c]
+	if ok {
+		goto Done
+	}
+	{
+		var m sync.Mutex
+		l = &m
+		bot.storeConvMap[c] = l
+	}
+
+Done:
+	bot.storeConvMapLock.Unlock()
+	(*l).Lock()
+	bot.storeWholeConvLock.RLock()
+	return func() {
+		(*l).Unlock()
+		bot.storeWholeConvLock.RUnlock()
+	}
 }
 
 func betabot() error {
@@ -73,7 +103,7 @@ func betabot() error {
 	rand.Seed(srand.MustSecure())
 
 	// init bot
-	var bot Bot
+	bot := &Bot{storeConvMap: make(map[*Conversation]*sync.Mutex)}
 
 	// init logger
 	{
@@ -197,7 +227,7 @@ func betabot() error {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			handledEvents := 0
+			var handledEvents uint
 			for {
 				gme, err := s.Recv()
 				if err != nil {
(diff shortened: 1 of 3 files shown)


go/cmd/betabot/main.go
```diff
@@ -206,26 +225,25 @@ func betabot() error {
...
-					// bot.logger.Debug("ignoring already handled event", zap.Any("event", gme))
+				if bot.isReplaying {
+					if gme.Event.Type == bertymessenger.StreamEvent_TypeListEnd {
+						bot.logger.Info("finished replaying logs from the previous sessions", zap.Uint("count", handledEvents))
+						bot.isReplaying = false
+					}
 					// replayed events
-					// noop
+					// bot.logger.Debug("ignoring already handled event", zap.Any("event", gme))
+					handledEvents++
```
moul, line 235: missing handledEvents if not in replay

go/cmd/betabot/main.go
```diff
@@ -93,11 +104,19 @@ func betabot() error {
...
 				return fmt.Errorf("parse %q: %w", *storePath, err)
 			}
 			bot.isReplaying = true // if the db exists on disk, then we switch the bot to replay mode
+			bot.store.StaffConvPK = pstore.StaffConvPK
+			i := len(pstore.Convs)
+			bot.store.Convs = make([]*Conversation, i)
+			for i != 0 {
+				i--
+				bot.store.Convs[i] = &Conversation{ConversationJSON: pstore.Convs[i]}
+			}
```
moul, line 119: I easily imagine there is a good reason to do this but I don't have enough knowledge in go to see what, can you add a quick comment explaining why you make an intermediate variable and manual assignment (step by step deepcopy) instead of just unmarshaling to the target?

go/cmd/betabot/main.go
Jorropo: > missing handledEvents if not in replay

@moul 
Why ? The only time it's red is in :
```go
bot.logger.Info("finished replaying logs from the previous sessions", zap.Uint("count", handledEvents))
```
When quitting replay mode.

go/cmd/betabot/main.go
```diff
@@ -58,13 +59,23 @@ type Bot struct {
 }
 
 type Conversation struct {
+	sync.Mutex
+	ConversationJSON
+}
+
+type ConversationJSON struct {
```
moul, line 66: can't you just put the sync.Mutex inside the Conversation?

go/cmd/betabot/main.go
Jorropo: > I easily imagine there is a good reason to do this but I don't have enough knowledge in go to see what, can you add a quick comment explaining why you make an intermediate variable and manual assignment (step by step deepcopy) instead of just unmarshaling to the target?

@moul I don't know it myself :D (Go vet hates {M,Unm}arshal of values with locker in them). (I'm still gonna add the comment.)

go/cmd/betabot/main.go
```diff
@@ -300,80 +319,94 @@ func (bot *Bot) handleUserMessageInteractionUpdated(ctx context.Context, _ *bert
...
 	if conv != nil && conv.IsOneToOne {
-		switch {
-		case conv.Count == 0 && checkValidationMessage(receivedMessage.GetBody()):
-			conv.Count++
-			bot.saveStore()
-			time.Sleep(1 * time.Second)
+		conv.Lock()
+		defer conv.Unlock()
+		switch conv.Step {
+		case 0:
```
moul, line 334: I like that you made the switch easier to read, but in that case I think we should directly go to a `map[int]func()`

go/cmd/betabot/main.go
Jorropo: Same as previous, no because I need an object without locker for {M,Unm}arshal due to go vet.

go/cmd/betabot/main.go
Jorropo: I see, do you think it would be worth to also make a iota instead of using 0,1 and 2 ?

go/cmd/betabot/main.go
moul: ok, so switch to a map of mutex

go/cmd/betabot/main.go
moul: yes, I forgot to log it elsewhere in the func but my goal was to always log this number

go/cmd/betabot/main.go
moul: iota will be dangerous here because we need to have step "names" however we store a "step number" in DB, so we need to stay consistent and just consider that "step 1 is after step 0" instead of "stepAddGroup is after stepAddContact"

but you can use a slice instead of a map :+1:

go/cmd/betabot/main.go
moul: for your information, seeing this PR making overkill change and increasing complexity just gives me the motivation to setup gorm right know

so please just focus on fixing the race (which is a real problem) instead of trying to make a too much smart store

Jorropo: > 
> 
> for your information, seeing this PR making overkill change and increasing complexity just gives me the motivation to setup gorm right know
> 
> so please just focus on fixing the race (which is a real problem) instead of trying to make a too much smart store

I'll do, but the store is still dumb AF, except there is 2, one with lockers for betabot usage and one without for json marshaling.

go/cmd/betabot/main.go
```diff
@@ -476,18 +551,31 @@ func (bot *Bot) interactUserMessage(ctx context.Context, body string, conversati
...
-	defer bot.storeMutex.Unlock()
-
-	// marshal
-	data, err := json.MarshalIndent(bot.store, "", "  ")
-	if err != nil {
-		panic(fmt.Errorf("marshal: %w", err))
-	}
+	var done bool
+	// Prevent concurrent saves
+	bot.storeOnce.Do(func() {
```
moul, line 556: what's the advantage of this VS a simple mutex?

go/cmd/betabot/main.go
```diff
@@ -300,83 +334,124 @@ func (bot *Bot) handleUserMessageInteractionUpdated(ctx context.Context, _ *bert
...
+		// auto-reply to user's messages
+		answer := getRandomReply()
+		if err := bot.interactUserMessage(ctx, answer, interaction.ConversationPublicKey); err != nil {
+			return fmt.Errorf("interact user message failed: %w", err)
+		}
+	}
+	return nil
+}
+
+var doSteps map[uint]func(context.Context, *Conversation, *Bot, *bertymessenger.AppMessage_UserMessage, *bertymessenger.Interaction, func()) (bool, error)
```
moul, line 363: move this global and the init function just before using it (before line 347)

in general, avoid `func init()` and `globals`

go/cmd/betabot/main.go
Jorropo: > what's the advantage of this VS a simple mutex?

Multiple call to save at the same time are merged as one.

go/cmd/betabot/main.go
Jorropo: > in general, avoid `func init()` and `globals`

I know but I don't know where to create the function map.

go/cmd/betabot/main.go
moul: that's a good idea, however, I think it's not working as expected:

* in a case, we can call it concurrently which may not guarantee that the object is the same, so we may just save the first call and lose the second one -> bad
* in the other case, we want to protect this by having a mutex to protect the base object, which makes this workflow useless

or I'm missing something?

go/cmd/betabot/main.go
moul: just before the call, before line 347

go/cmd/betabot/main.go
Jorropo: > in the other case, we want to protect this by having a mutex to protect the base object, which makes this workflow useless

That why it's called a last time at the closing of the program.

> In a case, we can call it concurrently which may not guarantee that the object is the same, so we may just save the first call and lose the second one -> bad

True, I was thinking this wasn't an issue because at worst this can't propagate very far (values are not lost they are not written to disk, but the next true save will write them).
I've fixed that by releasing the locks after the reset. (For this situation to happen again you would need to call save while editing the store (wich thx to some mutex is impossible and would just deadlocks)).

moul (approved):

PR #2420, opened 2020-10-05
chore: setup shake on forks

moul:
More info here: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target


moul:
since the goal of `pull_request_target` is to use the base branch's yml file, this PR won't run the iOS & Android builds even if the paths are correct

Jorropo (approved):
LGTM

PR #2423, opened 2020-10-06
fix(messenger): avoid common errors from handlers

glouvigny:


go/pkg/bertymessenger/handlers.go
```diff
@@ -721,7 +721,7 @@ func (h *eventHandler) interactionFetchRelations(tx *dbWrapper, i *Interaction)
 	} else {
 		i.Conversation = conversation
 
-		if i.Conversation.Type == Conversation_MultiMemberType {
+		if i.Conversation.Type == Conversation_MultiMemberType && i.MemberPublicKey != "" {
```
moul, line 724: is this a normal case?

glouvigny: absolutely not, this should be fetched after setting the `i.MemberPublicKey` value

aeddi (approved):
LGTM 👍

PR #2429, opened 2020-10-07
chore(ci): use expected commit at checkout for versionning

moul:
* `fetch-depth: 42` on `iOS` and `Android`, to increase chances of determining "pretty" version based on previous known tag
* `persist-credentials: false`, small security improvement
* `ref: ${{ github.event.pull_request.head.sha }}`, see https://github.com/actions/checkout/issues/299
* switch go versions to `1.15.x` instead of fixed patch version

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -26,11 +26,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.head.sha }}  # see https://github.com/actions/checkout/issues/299
 
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.15.2
+          go-version: 1.15.x
 
       - name: Setup Node
         uses: actions/setup-node@v1
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -26,11 +26,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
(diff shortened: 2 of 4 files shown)


.github/workflows/android.yml
```diff
@@ -26,11 +26,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 42                                 # fetch last 42 commits to increase chances of determining "pretty" version based on previous known tag
```
Jorropo, line 30:
```suggestion
          fetch-depth: 0                                  # fetch last 42 commits to increase chances of determining "pretty" version based on previous known tag
```
It's not like berty is a 1GB repo, I think it's worth downloading all the repo safe in case we merge more than 42 commits beetween the tag.

.github/workflows/ios.yml
```diff
@@ -26,11 +26,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 42                                 # fetch last 42 commits to increase chances of determining "pretty" version based on previous known tag
```
Jorropo, line 30: same

.github/workflows/release.yml
```diff
@@ -9,44 +9,37 @@ jobs:
     name: releaser
     runs-on: ubuntu-latest
     steps:
-      -
-        name: Checkout
+      - name: Checkout
```
Jorropo, line 12: I think you made typos.
Should be grouped together (with the new line right before the `-`) :
```yaml
- name: xxx
  uses: xxx

- name: yyy
  run: yyy
```

moul: I don't think so

.github/workflows/android.yml
moul: let's try

moul: yep, I confirm, there is only about 1 more second added :+1:

PR #2430, opened 2020-10-07
feat: Adding parameters for go-libp2p-tor-transport

Jorropo:
I'm experimenting with integration of the tor transport.
**This is WIP and will not protect any IP yet, I have some others things to do first, I've done many changes in `initutil` and I don't want more conflict when someone moves bridge's init to initutil.**

### DIFF
--- a/go/internal/initutil/ipfs.go
+++ b/go/internal/initutil/ipfs.go
@@ -17,6 +17,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/routing"
 	discovery "github.com/libp2p/go-libp2p-discovery"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	ma "github.com/multiformats/go-multiaddr"
 	"moul.io/srand"
 
 	"berty.tech/berty/v2/go/internal/config"
@@ -25,6 +26,9 @@ import (
 	"berty.tech/berty/v2/go/internal/tinder"
 	"berty.tech/berty/v2/go/pkg/bertyprotocol"
 	"berty.tech/berty/v2/go/pkg/errcode"
+	"berty.tech/berty/v2/go/pkg/tempdir"
+	tor "berty.tech/go-libp2p-tor-transport"
+	torcfg "berty.tech/go-libp2p-tor-transport/config"
 )
 
 func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
@@ -37,6 +41,10 @@ func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
 	fs.DurationVar(&m.Node.Protocol.MaxBackoff, "p2p.max-backoff", time.Minute, "maximum p2p backoff duration")
 	fs.BoolVar(&m.Node.Protocol.LocalDiscovery, "p2p.local-discovery", true, "local discovery")
 	fs.Var(&m.Node.Protocol.RdvpMaddrs, "p2p.rdvp", `list of rendezvous point maddr, ":dev:" will add the default devs servers, ":none:" will disable rdvp`)
+	fs.BoolVar(&m.Node.Protocol.Tor.Enabled, "tor.enable", defaultTorEnabliness, "if true tor will be enabled")
+	fs.StringVar(&m.Node.Protocol.Tor.BinaryPath, "tor.binary-path", "", "if set berty will use this external tor binary instead of his builtin one")
+	// FIXME: Create a custom command for help and listing about presets, they will quickly get crowded here.
+	fs.Var(&m.Node.Preset, "preset", `if set berty will use a preset instead of the default one. Curently only "anonymity" is available.`)
 }
 
 func (m *Manager) GetLocalIPFS() (ipfsutil.ExtendedCoreAPI, *ipfs_core.IpfsNode, error) {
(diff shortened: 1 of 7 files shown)


D4ryl00 (approved):
LGTM 👍

go/internal/initutil/manager.go
```diff
@@ -66,6 +66,14 @@ type Manager struct {
 			// The netry : `:none:` will disable all rdvp servers.
 			RdvpMaddrs flagStringSlice
 
+			// Tor
+			Tor struct {
+				Enabled       bool
+				BinaryPath    string // if "" that mean embedded must be used.
+				DontSetListen bool
+			}
+			AnonymityMode bool // Desactivate other stuff to enforce anonymity.
```
moul, line 75: I like the idea of adding this, but I prefer to use a `Preset string` instead of `Anonymity bool`, so we can have more than two presets

go/internal/initutil/ipfs.go
```diff
@@ -37,6 +40,10 @@ func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
 	fs.DurationVar(&m.Node.Protocol.MaxBackoff, "p2p.max-backoff", time.Minute, "maximum p2p backoff duration")
 	fs.BoolVar(&m.Node.Protocol.LocalDiscovery, "p2p.local-discovery", true, "local discovery")
 	fs.Var(&m.Node.Protocol.RdvpMaddrs, "p2p.rdvp", `list of rendezvous point maddr, ":dev:" will add the default devs servers, ":none:" will disable rdvp`)
+	fs.BoolVar(&m.Node.Protocol.Tor.Enabled, "tor.enable", false, "If true tor will be enabled.")
+	fs.StringVar(&m.Node.Protocol.Tor.BinaryPath, "tor.binary-path", "", "If set berty will use this external tor binary instead of his builtin one.")
+	fs.BoolVar(&m.Node.Protocol.Tor.DontSetListen, "tor.dont-set-listen", false, "If true tor will not auto set his listen address, allowing you to replace by an other one.")
+	fs.BoolVar(&m.Node.Protocol.AnonymityMode, "anonymity-force", false, `If true the node will be configurated to only use anonimous transports and will disable local discovery. Implies "-tor.enable=true"`)
```
moul, line 46:
```suggestion
	fs.BoolVar(&m.Node.Protocol.Tor.Enabled, "tor.enable", false, "if true tor will be enabled")
	fs.StringVar(&m.Node.Protocol.Tor.BinaryPath, "tor.binary-path", "", "if set berty will use this external tor binary instead of his builtin one")
	fs.BoolVar(&m.Node.Protocol.Tor.DontSetListen, "tor.dont-set-listen", false, "if true tor will not auto set his listen address, allowing you to replace by an other one")
	fs.BoolVar(&m.Node.Protocol.AnonymityMode, "anonymity-force", false, `if true the node will be configurated to only use anonimous transports and will disable local discovery. Implies "-tor.enable=true"`)
```

be consistent with the lines just above

go/internal/initutil/ipfs.go
```diff
@@ -37,6 +40,10 @@ func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
 	fs.DurationVar(&m.Node.Protocol.MaxBackoff, "p2p.max-backoff", time.Minute, "maximum p2p backoff duration")
 	fs.BoolVar(&m.Node.Protocol.LocalDiscovery, "p2p.local-discovery", true, "local discovery")
 	fs.Var(&m.Node.Protocol.RdvpMaddrs, "p2p.rdvp", `list of rendezvous point maddr, ":dev:" will add the default devs servers, ":none:" will disable rdvp`)
+	fs.BoolVar(&m.Node.Protocol.Tor.Enabled, "tor.enable", false, "If true tor will be enabled.")
+	fs.StringVar(&m.Node.Protocol.Tor.BinaryPath, "tor.binary-path", "", "If set berty will use this external tor binary instead of his builtin one.")
+	fs.BoolVar(&m.Node.Protocol.Tor.DontSetListen, "tor.dont-set-listen", false, "If true tor will not auto set his listen address, allowing you to replace by an other one.")
```
moul, line 45: I don't get how you want to use this

Why not just having something like:


```suggestion
	fs.StringVar(&m.Node.Protocol.Tor.Listen, "tor.listen", tor.NopMaddr3Str, "comment")
```

Jorropo: By default using `tor.enable` add the `tor.NopMAddr3Str` (`/onion/aaaaa...:1`) to the listen list.
Setting `DontSetListen` don't do that.

That mainly for debug purpose (or using onion 2), but I got a better idea, I'll remove `tor.dont-set-listen` and scan the listen list and if a `/onion{,3}/...` maddr is found the default tor listen address will not be added.

moul: yes but, your solution:
1. can only unset, but cannot set a custom value from the CLI
2. makes the `-h` not showing what is the default value

go/internal/initutil/manager.go
Jorropo: > I like the idea of adding this, but I prefer to use a `Preset string` instead of `Anonymity bool`, so we can have more than two presets

This is more than a preset, this sets `allowtcpdial` on tor, disable all other transports (except circuit) and disable single hop (not done yet).

aeddi (approved):
LGTM 👍

PR #2434, opened 2020-10-07
feat: add auto semver on ios

gfanton:
Add auto semver while running xcodegen:
Screenshot 2020-10-07 at 18 00 34

### DIFF
--- a/js/Makefile
+++ b/js/Makefile
@@ -1,6 +1,6 @@
 ## user variables
 
-VERSION ?= `cd ..; go run github.com/mdomke/git-semver`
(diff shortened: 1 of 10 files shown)


js/Makefile
```diff
@@ -1,6 +1,6 @@
 ## user variables
 
-VERSION ?= `cd ..; go run github.com/mdomke/git-semver`
+VERSION ?= `go run -mod=readonly -modfile=../go.mod github.com/mdomke/git-semver/v5`
```
moul, line 3: nice :+1:

moul (changes requested):
btw, there are other places in the code that use `go run` without the `/v5`, can you update them too?

moul (approved):

PR #2436, opened 2020-10-08
feat(betabot): Added /demo version

Jorropo:


clegirar (approved):
LGTM 👍

go/cmd/betabot/main.go
```diff
@@ -434,12 +436,25 @@ Type /help when you need info about available test commands! 📖`
 
 func doStep2(ctx context.Context, _ *Conversation, bot *Bot, receivedMessage *bertymessenger.AppMessage_UserMessage, interaction *bertymessenger.Interaction, unlock func()) (bool, error) {
 	unlock()
-	if receivedMessage.GetBody() == "/help" {
+	switch strings.ToLower(receivedMessage.GetBody()) {
```
clegirar, line 439: For the unavailable commands maybe can we put a default switch case that say `This command is not already available` or something like that ?
Now for example, a `/demo group` message, betabot auto reply a random answer

Jorropo: Very good point, I'll do that :)

clegirar:
LGTM 👍

PR #2447, opened 2020-10-09
feat: refactor account and contact group init

aeddi:
- allow the account group to be launched in local only mode (orbit-db will not sync store with other peers)
- send secrets and tag peers on account group
- force sending memberDevice and secrets during contact group init


D4ryl00 (approved):
LGTM 👍

glouvigny (approved):

PR #2448, opened 2020-10-09
fix: enable auto replicate on sent contact requests

glouvigny:


go/pkg/bertymessenger/api.go
```diff
@@ -297,21 +297,28 @@ func (svc *service) SendContactRequest(ctx context.Context, req *SendContactRequ
 		return nil, errcode.TODO.Wrap(err)
 	}
 
-	go svc.replicateContactGroupOnAllServers(req.BertyID.AccountPK)
+	go svc.autoReplicateContactGroupOnAllServers(req.BertyID.AccountPK)
```
n0izn0iz, line 300: Didn't see the goroutine here before, isn't this introducing races?

glouvigny: Should be fine, what race condition are you expecting there?

I just want to avoid failing requests on this call due to errors on this optional feature (and return a "success" value asap)

n0izn0iz: I dont expect something, I was asking to know if it was taken into consideration
There is an implicit (we should probably make it explicit) "contract" that only one thread will modify the `svc` by using a mutex in every handler and api call, this might escape the mutex lock of the call

n0izn0iz: But I think since it's already there it's outside the scope of this PR so this was just to trigger the discussion, we should open an issue propably

moul (approved):

PR #2449, opened 2020-10-09
chore: multidevices tests

D4ryl00:
Create tests for multidevices:
- TestMultiDevices_Basic

### DIFF
--- a/go/pkg/bertyprotocol/store_metadata_test.go
+++ b/go/pkg/bertyprotocol/store_metadata_test.go
@@ -656,3 +656,141 @@ func TestMetadataGroupsLifecycle(t *testing.T) {
 	require.Equal(t, groups[0].SecretSig, g2.SecretSig)
 	require.Equal(t, groups[0].GroupType, g2.GroupType)
 }
+
+func TestMultiDevices_Basic(t *testing.T) {
+	testutil.FilterStabilityAndSpeed(t, testutil.Unstable, testutil.Slow)
+
+	memberCount := 2
+	deviceCount := 3
+	totalDevices := memberCount * deviceCount
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	peers, _, cleanup := createPeersWithGroup(ctx, t, "/tmp/multidevices_test", memberCount, deviceCount)
+	defer cleanup()
+	// make peer index
+	pi := [][]int{}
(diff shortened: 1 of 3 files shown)


moul (approved):
LGTM, can you rebase your commits?

D4ryl00:
> LGTM, can you rebase your commits?

done

go/pkg/bertyprotocol/store_metadata_test.go
```diff
@@ -656,3 +656,141 @@ func TestMetadataGroupsLifecycle(t *testing.T) {
...
+	testutil.FilterStabilityAndSpeed(t, testutil.Unstable, testutil.Slow)
+
+	memberCount := 2
+	deviceCount := 3
+	totalDevices := memberCount * deviceCount
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	peers, _, cleanup := createPeersWithGroup(ctx, t, "/tmp/multidevices_test", memberCount, deviceCount)
```
Jorropo, line 670:
```suggestion
	tDir, err := ioutil.TempFile("/tmp", "multidevices_test")
	defer os.Remove(tDir.Name())
	require.NoError(t, err)
	peers, _, cleanup := createPeersWithGroup(ctx, t, tDir.Name(), memberCount, deviceCount)
```

Jorropo (changes requested):

go/pkg/bertyprotocol/store_metadata_test.go
D4ryl00: I did as other tests in this file, but `createPeersWithGroup` doesn't use the path currently. I don't know why it's not used, but if we must create temp file, I think it should be done in the helper function `createPeersWithGroup`, not in the test directly. But it's an other problem ;)
So I think that I did is good to be homogeneous with other tests.

Jorropo: > 
> 
> I did as other tests in this file, but `createPeersWithGroup` doesn't use the path currently. I don't know why it's not used, but if we must create temp file, I think it should be done in the helper function `createPeersWithGroup`, not in the test directly. But it's an other problem ;)
> So I think that I did is good to be homogeneous with other tests.

As you wish, I just don't really like static path (I guess this is still mergeable BTW)

moul (approved):
LGTM

PR #2455, opened 2020-10-09
ci: Cancel previous Runs to avoid running out of runners.

Jorropo:


.github/workflows/go.yml
```diff
@@ -24,6 +24,11 @@ jobs:
     runs-on: ubuntu-latest
     needs: golangci-lint
```
moul, line 25: this rule should be added on golangci-lint (first rule)

.github/workflows/push.yml
```diff
@@ -0,0 +1,27 @@
...
+          fetch-depth: 1
+          persist-credentials: false
+          ref: ${{ github.event.pull_request.head.sha }}  # see https://github.com/actions/checkout/issues/299
+
+      - name: Resolve IDS
+        id: resolve-ids
+        run: ./tool/build/shell/resolve-workflows-for-cancel.py
+
+      - name: Cancel Workflows
+        if: github.ref != 'refs/heads/master'
```
moul, line 23: can you move this if at the job level instead of the step level?

Jorropo: > can you move this if at the job level instead of the step level?

@moul 
I don't think (I've red the github action syntaxe sheet and there is no workflow `if` key, only on a per job basis. (plus this whole workflow doesn't run for more than 10s so I don't think it's a waste that big.)).

moul: that's strange -> https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

Jorropo: > 
> 
> that's strange -> [docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif)

@moul 
Ah yes my bad, I havn't though of that

moul (approved):

PR #2468, opened 2020-10-12
fix: add user friendly error message on invitation error

ekelen:
Even if we solve the current issue, it makes sense to have a user-friendly error message in place somewhere.

image


n0izn0iz (approved):
code lgtm

clegirar:
Yes i agree but it's a little weird i think to have this error message and to have the conversation in the home no ?

clegirar:
But lgtm for the code too 👍

ekelen:
> Yes i agree but it's a little weird i think to have this error message and to have the conversation in the home no ?

Yes you're right, I will change the error text

PR #2477, opened 2020-10-12
feat: initial version of the PeerList call + new 'berty peers' CLI command

Jorropo:


api/bertytypes.proto
```diff
@@ -985,3 +985,11 @@ message SystemInfo {
     string system_username = 28;
   }
 }
+
+message Libp2pPeers {
+  message Request {}
+  message Reply {
+    // List of maddrs.
+    string repeated = 1;
```
moul, line 993:
```suggestion
    repeated string addrs = 1;
```

api/bertytypes.proto
```diff
@@ -985,3 +985,11 @@ message SystemInfo {
     string system_username = 28;
   }
 }
+
+message Libp2pPeers {
```
moul, line 989: btw, if we can't have more data than what libp2p can return, yes, name this `libp2p-peers`, but as soon as we know that we will enrich this data with additional metadata, this call should be named `p2p-peers` and the peers should be a struct containing the address, but also additional data

Jorropo:
@moul I think you should fixup my commit, it's useless.

moul (approved):

PR #2482, opened 2020-10-13
WIP: feat: Single hop tor for compatiblity mode

Jorropo:
This mode uses a single tor hop if we are running in compatible mode (with other transports open).

Currently this PR is stalled (require a PR to allow client services while running single hop on Tor).
This is currently disabled by tor for reasons. It's probably because that would cause issue with people using exit relays with a single hop connection (it's disabled and fail at the exit relay level) but for Tor 2 Tor connection it's perfectly fine.

PS: This can be made working by manually hardcoding the default tunnel length value at `1` but that would disable usage of tor for anonymity.


Jorropo:
Archived, stuck right now.

PR #2494, opened 2020-10-15
fix(messenger): in conv keep sent state only on last message

clegirar:
Fix https://github.com/orgs/berty/projects/42#card-47349108 https://github.com/orgs/berty/projects/42#card-47342612 https://github.com/orgs/berty/projects/42#card-47342467


zxxma:
Ideally, it could be perfect to let just the arrow, on others message, but 👍 LGTM!

PR #2496, opened 2020-10-15
fix: Improve MC stability

D4ryl00:
Stabilize MC: close connections when the driver lost the connection with the remote peer

Rename the transporter with a name more generic: in the future, we should reutilize the transporter for multiple drivers (BLE, Nearby)

### DIFF
--- a/go/internal/multipeer-connectivity-transport/transport.go
+++ b/go/internal/multipeer-connectivity-transport/transport.go
@@ -20,63 +20,61 @@ const DefaultBind = "/mc/Qmeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
 
 // logger is global because HandleFoundPeer must be able to call it
 // FIXME: remove global logger
-var logger *zap.Logger = zap.L().Named("mc-transport")
+var logger *zap.Logger = zap.NewNop()
 
 // Transport is a tpt.transport.
-var _ tpt.Transport = &Transport{}
+var _ tpt.Transport = &ProximityTransport{}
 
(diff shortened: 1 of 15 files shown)


go/internal/multipeer-connectivity-transport/transport.go
```diff
@@ -20,63 +20,61 @@ const DefaultBind = "/mc/Qmeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
...
 
 // Transport is a tpt.transport.
-var _ tpt.Transport = &Transport{}
+var _ tpt.Transport = &ProximityTransport{}
 
-// Transport represents any device by which you can connect to and accept
+// ProximityTransport represents any device by which you can connect to and accept
 // connections from other peers.
-type Transport struct {
+type ProximityTransport struct {
```
Jorropo, line 30: (unimportant)
Are you sure about this name ?
I mean I don't think it's gonna be the only proximitve transport (ex: lora).
Wouldn't be better to have a name like `BleWifiTransport` ?

Jorropo (approved):
SGTM

go/internal/multipeer-connectivity-transport/transport.go
D4ryl00: I don't found a better name ^^ BleWifi no because MC is not BLE, and this transporter will be the same for Nearby (android).

Jorropo: > I don't found a better name

Yes same for me :) (this name fine, being not totally accurate is not a problem).

PR #2499, opened 2020-10-16
feat(bridge): use initutil to initialize the mobile bridge

moul:
* [x] use `internal/initutil` to initialize the bridge
* [x] cleanup: have the minimal exposed API (with current usage from mobile)
* [x] clean: merge Bridge, MessengerBridge into a unique type
* [x] port tests
* [x] remove the tracer on mobile
* [x] remove debug counter push-notif
* [x] port js&native code
* [x] check if it runs on yolo

Replaces #2491
Depends on #2500

PR #2500, opened 2020-10-16
Remove js monorepo

n0izn0iz:
blocks #2499


js/android/app/src/main/AndroidManifest.xml
```diff
@@ -2,9 +2,10 @@
   package="tech.berty.app">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.VIBRATE"/>
     <uses-permission android:name="android.permission.CAMERA" />
+
     <application
-      android:networkSecurityConfig="@xml/network_security_config"
```
aeddi, line 7: You should not remove this or the webui won't work anymore.
I think the same problem will occur for the tcp_reverse w/ metro on localhost.

js/android/app/src/main/AndroidManifest.xml
```diff
@@ -46,12 +47,6 @@
             <data android:scheme="berty" />
         </intent-filter>
       </activity>
-      <meta-data
-        android:name="com.shakebugs.APIClientID"
-        android:value="${SHAKE_API_ID}" />
-      <meta-data
-        android:name="com.shakebugs.APIClientSecret"
-        android:value="${SHAKE_API_SECRET}" />
```
aeddi, line 54: Are you sure about this?

js/packages/berty-app/android/build.gradle
```diff
@@ -1,68 +0,0 @@
...
-}
-
-ext.getEnvValue = { key ->
-    if (System.getenv(key)) { // Get var from system env first
-        return System.getenv(key)
-    } else if (System.getProperty(key)) { // if null, get var from env file
-        return System.getProperty(key)
-    }
-    return "" // if null, return an empty string
-}
```
aeddi, line 68: You should not remove this nor the different parts of the config using it

aeddi (changes requested):

js/android/app/src/main/AndroidManifest.xml
n0izn0iz: Metro access is in a very bad shape on master, on this branch it's automatic, doesn't even need the reverse I think

js/android/app/src/main/AndroidManifest.xml
n0izn0iz: No I need to add it back, I left android for now since we focus on iOS and we need to merge this asap

js/Makefile
```diff
@@ -389,32 +355,14 @@ packages/go-bridge/android/libs/gobridge.aar: $(bridge_src)
...
 
-packages/berty-app/android/app/debug.keystore:
-	cp packages/berty-app/node_modules/react-native/template/android/app/debug.keystore $@
-
-packages/storybook-mobile/android/app/debug.keystore:
-	cp packages/berty-app/node_modules/react-native/template/android/app/debug.keystore $@
-
-packages/berty-app/node_modules/react-native/react.gradle.orig: packages/berty-app/patch/react.gradle.patch
-	@echo "patching file $(patsubst %.orig, %, $@)"
-	@patch -f $(patsubst %.orig, %, $@) $< 1>/dev/null || true
```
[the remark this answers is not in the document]
n0izn0iz: I don't know but it looks like a hack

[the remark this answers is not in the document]
aeddi: I think it was made to disable the debug related features on the different release targets we have.
The patch replace something like: `if target == release` by `if target != debug` because we have 3 differents release targets.
It's probably not the best way to do it but now with this change, all the debug features will be enabled on the different release targets.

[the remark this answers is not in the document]
aeddi: I don't know, maybe. I didn't do the initial setup. I'll see what to do once it's merge, after thinking about it, it won't hurt to start from scratch. ;)

PR #2503, opened 2020-10-16
feat: implement native tempdir getter for Android and iOS

aeddi:
We should pass the result of the `getCacheDir().getAbsolutePath()` Java method to `SetAndroidCacheDir(path)`.
Could be done in #2499 :) (or I will do it when it will be merged)


moul (approved):

Jorropo (approved):
LGTM

PR #2507, opened 2020-10-16
WIP: build(messenger): allow iOS dev landscape orientation

ekelen:
I think this should work, but I'm not sure. The idea is to lock portrait orientation on iOS everywhere, except for Debug and Yolo releases on **iPad**, where we also need to support landscape.

@gfanton I would much appreciate your input. I added an iPad-specific key `UISupportedInterfaceOrientations~ipad` and corresponding `INTERFACE_ORIENTATION_IPAD` in the config, but for some reason my iPhone is now allowing landscape, not just iPad.


gfanton:
fixed on #2751

PR #2510, opened 2020-10-19
fix: temporally force private reachability

gfanton:


go/internal/config/berty.go
```diff
@@ -59,8 +59,8 @@ var BertyMobile = &BertyConfig{
 	},
 	DefaultSwarmAddrs: []string{
 		mc.DefaultBind,
-		"/ip4/0.0.0.0/tcp/0",
-		"/ip6/::/tcp/0",
+		// "/ip4/0.0.0.0/tcp/0",
+		// "/ip6/::/tcp/0",
```
Jorropo, line 63: Why ?

PR #2519, opened 2020-10-20
WIP: feat(messenger): add devtool Network Map

clegirar:


js/packages/store/provider.tsx
```diff
@@ -256,9 +260,11 @@ export const MsgrProvider: React.FC<any> = ({ children, daemonAddress, embedded
 		}
 
 		const messengerClient = Service(messengerpb.MessengerService, rpc, messengerMiddlewares)
+		const protocolClient = Service(protocolpb.ProtocolService, rpc, null)
 
 		dispatch({ type: 'CLEAR' })
 		dispatch({ type: 'SET_CLIENT', payload: { client: messengerClient } })
+		dispatch({ type: 'SET_PROTOCOL_CLIENT', payload: { client: protocolClient } })
```
n0izn0iz, line 267: please merge the actions or add a check for the protocolClient in gates

moul:
the network map should not be located in devtools but in the main menu

clegirar:
I think just 1 point can be done in another PR (except if we have a design before):
- maybe it's just not implemented, but the dropdown menu at the right of a peer does nothing when I click on it
But okay for anothers improvements 👍

moul (approved):

js/packages/components/settings/NetworkMap.js
```diff
@@ -0,0 +1,131 @@
...
+	const prevPeers = usePrevious(sortPeers)
+
+	useEffect(() => {
+		if (!called || refresh) {
+			if (refresh) {
+				setRefresh(false)
+			}
+			call()
+		}
+	}, [called, call, refresh, setRefresh])
```
n0izn0iz, line 61: you can avoid this whole block by moving the method hook in the parent component and passing `call` as a prop directly

clegirar: yes it's part the part of code that i'm not proud but i didn't know how to do that properly but okay 👍

js/packages/components/settings/NetworkMap.js
```diff
@@ -112,6 +102,17 @@ export const NetworkMap = () => {
...
+
```
n0izn0iz: now you can remove setRefresh and the effect entirely and use call directly

moul:
can you squash some commits?

clegirar:
Yes i'll do that 👍

PR #2538, opened 2020-10-21
WIP: Translate go CLI

Jorropo:


go/internal/initutil/ipfs.go
```diff
@@ -32,9 +32,9 @@ import (
 )
 
 func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
-	fs.StringVar(&m.Node.Protocol.IPFSListeners, "p2p.ipfs-listeners", "/ip4/0.0.0.0/tcp/0,/ip4/0.0.0.0/udp/0/quic", "IPFS listeners")
-	fs.StringVar(&m.Node.Protocol.IPFSAPIListeners, "p2p.ipfs-api-listeners", "", "IPFS API listeners")
-	fs.StringVar(&m.Node.Protocol.Announce, "p2p.ipfs-announce", "", "IPFS announce addrs")
```
moul, line 37: the API is ugly to use in go, we need something like `t("cli.ipfs.listeners")`

go/internal/i18n/locate_lang.go
```diff
@@ -0,0 +1,16 @@
+package main
```
moul, line 1: no `main` package in `go/internal`

go/internal/i18n/generate-json.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
```
moul, line 1: avoid python if possible, and especially in the `go/` dir

.github/workflows/go-i18n.yml
```diff
@@ -0,0 +1,35 @@
+name: Go i18n
```
moul, line 1: avoid creating new workflows, write new jobs in existing ones instead

moul (changes requested):
see review comments

btw, I'm for closing this since it's ultra low-priority and adding useless complexity during development

go/internal/i18n/locate_lang.go
Jorropo: Yes sorry, I was debbuging (it's not ready yet).

go/internal/i18n/generate-json.py
Jorropo: Can I move it in a folder under `/tool/` ?

.github/workflows/go-i18n.yml
Jorropo: That was for having a custom path trigger, this workflows need to be runned only if some touches `*/i18n.go`.

PR #2548, opened 2020-10-22
WIP chore: improve CLI

moul:


go/internal/initutil/ipfs.go
```diff
@@ -32,19 +33,44 @@ import (
...
 func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
-	fs.Var(&m.Node.Protocol.IPFSListeners, "p2p.ipfs-listeners", "IPFS listeners")
-	fs.Var(&m.Node.Protocol.IPFSAPIListeners, "p2p.ipfs-api-listeners", "IPFS API listeners")
-	fs.Var(&m.Node.Protocol.Announce, "p2p.ipfs-announce", "IPFS announce addrs")
-	fs.Var(&m.Node.Protocol.NoAnnounce, "p2p.ipfs-no-announce", "IPFS exclude announce addrs")
+	m.SetupPresetFlags(fs)
+	fs.StringVar(&m.Node.Protocol.SwarmListeners, "p2p.swarm-listeners", ":default:", "IPFS swarm listeners")
+	fs.StringVar(&m.Node.Protocol.IPFSAPIListeners, "p2p.ipfs-api-listeners", "", "IPFS API listeners")
+	fs.StringVar(&m.Node.Protocol.Announce, "p2p.ipfs-announce", "", "IPFS announce addrs")
+	fs.StringVar(&m.Node.Protocol.NoAnnounce, "p2p.ipfs-no-announce", "", "IPFS exclude announce addrs")
```
Jorropo, line 40: Any particular reason to change to comma separated value ?
The previous option were in my opinion better, you could either use comma separated value or set the flag multiple times (and mix them).

go/internal/initutil/ipfs.go
```diff
@@ -82,105 +111,78 @@ func (m *Manager) getLocalIPFS() (ipfsutil.ExtendedCoreAPI, *ipfs_core.IpfsNode,
...
-				} else {
-					torOpts = torcfg.Merge(torOpts, torcfg.SetBinaryPath(m.Node.Protocol.Tor.BinaryPath))
-				}
 
-				if setListenTor {
-					swarmAddrs = append(swarmAddrs, tor.NopMaddr3Str)
-				}
+			if !hasTorMaddr(swarmAddrs) {
+				swarmAddrs = append(swarmAddrs, tor.NopMaddr3Str)
+			}
```
Jorropo, line 166:
```suggestion
```
Not required anymore as this is set in the `ipfsConfigPatch` earlier.

go/internal/initutil/ipfs.go
moul: the previous version has two major issues:

1. there was not default value printed in the `-h`, just `[]`
2. there was too much magic, even if the `-h` says `[]`, there will be default values

The current version is a way better since it always reflect what has been set exactly or use the exact same default value indicated in `-h`

moul: + ultra strange behavior when mixing values from env + values from CLI with no easy way to "override" with one expected value

go/internal/initutil/ipfs.go
moul: the ipfsconfigpatch is only used for tor==required, I will try to refactor the code to remove this

PR #2552, opened 2020-10-22
fix: implem I18n wrapper

clegirar:


moul:
two things to do related to this PR:
1. when you add new translatable texts, run `make generate` in the `js/` folder, you will maybe need to `rm gen.sum` first; it will regenerate the fake "debug language" which replaces every translation with its corresponding XPath
2. we need to easily be able to switch from a language to another from the app (in a devtool)

Then, yes, it will be a good thing to be more and more exhaustive in the translatable ratio :+1:

PR #2553, opened 2020-10-22
chore: small makefile refactor to handle relative go.mod

moul:
* fix: restore version in systeminfo
* improve maintainability

Depends on #2555

### DIFF
--- a/js/Makefile
+++ b/js/Makefile
@@ -1,6 +1,6 @@
 ## User variables
 
-VERSION ?= `go run -mod=readonly -modfile=../go.mod github.com/mdomke/git-semver/v5`
+VERSION ?= `go run -mod=readonly -modfile=$(PWD)/../go.mod github.com/mdomke/git-semver/v5`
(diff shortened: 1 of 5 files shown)


js/Makefile
```diff
@@ -353,3 +353,8 @@ $(pkg_mod): $(pkg_desc)
 	mkdir -p $(pkg_mod)
 	touch $(pkg_mod)
 	touch node_modules/go.mod # avoid go to lookup in this directorys
+
+#### Debug
+
+print-%:
+	@echo $* = $($*)
```
aeddi, line 360: Do you want to keep it or did you forget to remove it?

aeddi (approved):
LGTM 👍

js/Makefile
moul: I want to keep it, I often use it for debugging, basically, each time we had issues with Makefile in the past, I used it and asked other people to run various `make print-Something` commands

PR #2555, opened 2020-10-23
feat: cleanup / refactor app projects

### LINKED ISSUE #2520: Fix android build
see @aeddi comments here: https://github.com/berty/berty/pull/2500

Done in #2555

aeddi:
- move ios specific files (vendor, Gemfile, etc...) to ios folder
- build correct version of Android on CI (Yolo instead of store release)
- strip armv7 and armv7s archs on iOS CI build (~87Mo -> ~55Mo)
- Makefile:
  - cleanup unused variables and rules
  - add a protection that output an error if make is not exec from Makefile dir
  - several fixes
- restore Staff and Yolo variants on Android
- rename Android app ID `tech.berty.app` -> `tech.berty.android` (to be consistent with `tech.berty.ios`)
- fixes crashlytics and shakebugs on Android and iOS
- restore Android network security config to access to the webUI
- remove excluded archs / enable active arch only on pods and Xcode project
- fix pods project warnings (swift 4.x -> 5.0, iOS deployment == 12)

fix #2520


js/Makefile
```diff
@@ -173,23 +186,32 @@ ios.gomobile: ios/Frameworks/Bertybridge.framework
...
-android.app_deps: node_mod android.gomobile android/app/debug.keystore
+android.app_deps: node_modules android.gomobile
 
 .PHONY: android.debug
 android.debug: android.app_deps
-	$(eval CURRENT_JAVA_VER ?= `java -version 2>&1 | sed -n ';s/.* version "\(.*\)\.\(.*\)\..*"/\1\2/p;'`)
+	$(eval CURRENT_JAVA_VER ?= $(shell java -version 2>&1 | sed -n ';s/.* version "\(.*\)\.\(.*\)\..*"/\1\2/p;'))
 	@if [ "$(CURRENT_JAVA_VER)" != "$(required_java_ver)" ]; then \
 	  >&2 echo "ERROR: Invalid java version, want $(required_java_ver), got $(CURRENT_JAVA_VER)"; exit 1; \
 	fi
```
moul, line 196: you should make a `_check_java` rule

js/Makefile
```diff
@@ -201,16 +223,16 @@ android.log:
 .PHONY: android.emulator
 android.emulator:
 	$(call check-program, adb emulator)
-	$(eval ANDROID_EMULATOR ?= `emulator -list-avds | head -n1`)
+	$(eval ANDROID_EMULATOR ?= $(shell emulator -list-avds | head -n1))
```
moul, line 226: I'm not 100% sure, but with my current understanding on how makefile works, I think this will be called even if we manually set the var at call time

js/Makefile
```diff
@@ -201,16 +223,16 @@ android.log:
...
 	  >&2 echo "ERROR: no Android emulator AVD found"; exit 1; \
 	fi
-	emulator @$(ANDROID_EMULATOR) -logcat *:i
+	emulator @$(ANDROID_EMULATOR) -logcat *:i -no-snapshot-save
 
 .PHONY: android.reverse_tcp
 android.reverse_tcp:
 	$(call check-program, adb)
-	$(eval ANDROID_DEVICE ?= `adb devices | tail +2 | head -1 | cut -f 1`)
+	$(if $(ANDROID_DEVICE),,$(eval ANDROID_DEVICE = $(shell adb devices | tail +2 | head -1 | cut -f 1)))
```
moul, line 235:
```suggestion
	$(if $(ANDROID_DEVICE),,$(eval ANDROID_DEVICE ?= $(shell adb devices | tail +2 | head -1 | cut -f 1)))
```

nope?

js/Makefile
```diff
@@ -342,14 +359,9 @@ android/libs/gobridge.aar: $(bridge_src)
 		./go/framework/bertybridge
 	touch $@
 
-android/app/debug.keystore:
-	cp node_modules/react-native/template/android/app/debug.keystore $@
-
 #### React
 
-$(pkg_mod): $(pkg_desc)
+node_modules: package.json
```
n0izn0iz, line 364: the dependency should be yarn.lock I think

js/Makefile
aeddi: Ok done 👍 I'll push it in few minutes

js/Makefile
aeddi: This rule aims to easily launch the first Android emulator found in the list of those available on the machine (basically: list the AVDs, launch the first one).
So the var ANDROID_EMULATOR is not intended to be set by the user of the Makefile, so I'll pass it in lowercase to make it clearer.

js/Makefile
aeddi: I just tested and both syntaxes work: if the variable is set by the user, it is taken into account, otherwise this line lists the devices and sets the variable.

js/Makefile
aeddi: The previous rule was only based on `{node_modules,packages/*/nodemodules}: {package.json,packages/*/packages.json}` and it looks ok for me. But I can add `node_modules: package.json yarn.lock` if you want to ;)

n0izn0iz: Didn't that into account that if we only have `yarn.lock`, editing the `package.json` and running the make rules won't run `yarn install`, you're right, I'm for `package.json` only then

js/Makefile
moul: Oh yes, sorry I was trying to fix your line but didn't get it entirely

If your change has the same effect we had previously, I strongly prefer the original way

moul:
```suggestion
	$(eval ANDROID_DEVICE ?= `adb devices | tail +2 | head -1 | cut -f 1`)
```

aeddi: Tested this line, it doesn't work.
```
eval_test:
	$(eval ANDROID_DEVICE ?= `adb devices | tail +2 | head -1 | cut -f 1`)
	echo $(ANDROID_DEVICE)
```

output:
```
echo
```

And without the `?`:
```
eval_test:
	$(eval ANDROID_DEVICE ?= `adb devices | tail +2 | head -1 | cut -f 1`)
	echo $(ANDROID_DEVICE)
```

output:
```
echo `adb devices | tail +2 | head -1 | cut -f 1`
emulator-5554
```

But with the 2nd option (without `?`), it will ignore the value set by the user.

moul: ok, that's probably due do the `ANDROID_DEVICE ?=` at the beginning of the file

I dislike this complicated Makefile style but if it fixes real problem, up to you to merge like this :+1:

moul: fyi, my favorite method is just:

```make
ANDROID_DEVICE ?= `adb devices | tail +2 | head -1 | cut -f 1`
```

at the beginning of the file, it will not assign the result of `adb devices` but just the shell command to run only when it will really be called

as soon as we have `eval` in Makefile, we reduce the maintainability

an alternative I like is having only one helper function that centralizes all the complicated stuff, so you can just depend on that helper to do the eval(triple(condition(depending(on(100(lines(above)))))))` which makes it easier to understand at the end

aeddi: Ok got it but if we do this:
```
ANDROID_DEVICE ?= `adb devices | tail +2 | head -1 | cut -f 1`
```

We will overide the behavior of `react-native run-android` that automatically create and launch a temporary emulator (even if no AVD is available) in `android.debug` and `android.release` since `ANDROID_DEVICE` will always be set. I'll keep it as it is.

PR #2557, opened 2020-10-24
chore(js): bump typescript

n0izn0iz:


n0izn0iz:
closing since react-native will cut a release very soon and it will probably be easier to upgrade everything at this point, maybe everything will already be done properly by the typescript template even

PR #2628, opened 2020-11-02
feat: Add temporary Hacked Relay

Jorropo:


go/internal/initutil/ipfs.go
```diff
@@ -46,6 +46,7 @@ func (m *Manager) SetupLocalIPFSFlags(fs *flag.FlagSet) {
 	fs.StringVar(&m.Node.Protocol.Tor.Mode, "tor.mode", defaultTorMode, "if true tor will be enabled")
 	fs.StringVar(&m.Node.Protocol.Tor.BinaryPath, "tor.binary-path", "", "if set berty will use this external tor binary instead of his builtin one")
 	fs.BoolVar(&m.Node.Protocol.DisableIPFSNetwork, "p2p.disable-ipfs-network", false, "disable as much networking feature as possible, useful during development")
+	fs.BoolVar(&m.Node.Protocol.RelayHack, "p2p.relay-hack", false, "if set berty will force libp2p to use randoms relays from the config and will advertise as ours optimisticaly")
```
moul, line 49:
```suggestion
	fs.BoolVar(&m.Node.Protocol.RelayHack, "p2p.relay-hack", false, "*temporary flag*; if set, Berty will only advertise relayed addresses")
```

go/internal/initutil/ipfs.go
```diff
@@ -181,8 +182,58 @@ func (m *Manager) getLocalIPFS() (ipfsutil.ExtendedCoreAPI, *ipfs_core.IpfsNode,
...
+				libp2p.Transport(proximity.NewTransport(m.ctx, logger, mc.NewDriver(logger))),
+			)
+		}
+
+		if m.Node.Protocol.RelayHack {
+			// Resolving addresses
+			pis, err := ipfsutil.ParseAndResolveRdvpMaddrs(m.GetContext(), m.initLogger, []string{
+				"/ip4/51.159.21.214/udp/4040/quic/p2p/QmdT7AmhhnbuwvCpa5PH1ySK9HJVB82jr3fo1bxMxBPW6p",
+				"/ip4/51.15.25.224/udp/4040/quic/p2p/12D3KooWHhDBv6DJJ4XDWjzEXq6sVNEs6VuxsV1WyBBEhPENHzcZ",
+				"/ip4/51.75.127.200/udp/4141/quic/p2p/12D3KooWPwRwwKatdy5yzRVCYPHib3fntYgbFB4nqrJPHWAqXD7z",
```
moul, line 195: can you add them in the config?

Jorropo: > can you add them in the config?

I've tried but that also require to update bertytypes, I'm scared to break somethings in the future at rollback.

go/internal/initutil/ipfs.go
Jorropo: > Berty will only advertise relayed addresses

It will not stop advertising other addresses (we want to have as much chances as possible).
Maybe `*temporary flag*; shouldn't be set` would be good ?

go/internal/initutil/ipfs.go
Jorropo: My bad, `bertytypes` is not a protobuf generated package.

go/framework/bertybridge/config.go
```diff
@@ -30,6 +30,9 @@ func (c *Config) AppendCLIArg(arg string)                         { c.cliArgs =
 func (c *Config) manager() (*initutil.Manager, func(), error) {
 	manager := initutil.Manager{}
 
+	// FIXME: Disable this option once the discovery is fixed.
+	manager.Node.Protocol.RelayHack = true
```
moul, line 34: take the habit to update the js code, else we will have some rules here, some other in the other place

config/config.yml
```diff
@@ -34,3 +34,7 @@ p2p:
     # jorropo
     - maddr: "/ip4/51.75.127.200/tcp/4141/p2p/12D3KooWPwRwwKatdy5yzRVCYPHib3fntYgbFB4nqrJPHWAqXD7z"
     - maddr: "/ip4/51.75.127.200/udp/4141/quic/p2p/12D3KooWPwRwwKatdy5yzRVCYPHib3fntYgbFB4nqrJPHWAqXD7z"
+  relayHack:
```
moul, line 37: what about just naming it `relays`, since they can be used in other valid circumstances too?

PR #2633, opened 2020-11-02
fix: Correctly setup RelayHack on mobile

Jorropo:


go/framework/bertybridge/config.go
```diff
@@ -49,6 +46,9 @@ func (c *Config) manager() (*initutil.Manager, func(), error) {
 		return nil, nil, fmt.Errorf("invalid CLI args, should only have flags")
 	}
 
+	// FIXME: Disable this option once the discovery is fixed.
+	manager.Node.Protocol.RelayHack = true
```
moul, line 50: this is still incorrect, you need to call the --relay-hack=true from JS

Jorropo: I know :'(

moul (approved):

PR #2644, opened 2020-11-03
fix: android flavors config for network security and app name

aeddi:
- Remove network-security-config from debug variant to restore automatic binding between Android device/emulator and metro server
- Set different app name for each variant


n0izn0iz (approved):
lgtm

PR #2663, opened 2020-11-05
fix: removed some dev warnings in rn client

glouvigny:
Other identified warnings but not covered here:

- When displaying a notification, fixed in the lib, but no released has been made since https://github.com/robcalcroft/react-native-in-app-notification/pull/43
- When displaying search results due to a scroll view hosting a list view


n0izn0iz (approved):
code lgtm

clegirar:
LGTM 👍

PR #2675, opened 2020-11-08
fix: new i18n updates from crowdin

berty-assistant:


js/packages/berty-i18n/locale/en/messages.json
```diff
@@ -250,7 +250,7 @@
         "title": "Share ID on discord",
         "alert-title": "Do you want to share your QR code to \"#dev-logs\" on Berty Discord?",
         "alert-message": "This will allow other staff to scan your QR code or copy your contact link.",
-        "alert-accept-button": "Yep \uD83D\uDC4D",
+        "alert-accept-button": "Yep 👍",
```
moul, line 253: @clegirar, crowdin automatically replaces unicodes with their values

do you confirm that it will work? else we need to find a solution, i.e., updating the library to support unicodes like this, or have some replacements like: `:+1:` -> `👍`

clegirar: In the simu it works, i'll test on my iPod 👍

clegirar: It works on device (iPod)

moul: Ok thx!

moul (approved):

PR #2684, opened 2020-11-09
fix: go's `make install` on darwin

Jorropo:
`github.com/zcalusic/sysinfo` calls unavailable syscalls on darwin.


moul (approved):

D4ryl00 (approved):

moul (changes requested):
strange -> the go workflow was not run even if we have both .go files and the workflow file that changed

PR #2692, opened 2020-11-10
feat: media sharing

n0izn0iz:
remaining TODO
- [x] prevent late-loading image from overriding source
- [x] Prevent from uploading images too big
- [x] Finish EditProfile ui
- [x] send user info when joining 1to1
- [x] add test
- [x] Clean/Refracto
- [x] Review
- [ ] ~Party~ Party canceled, I dedicate this to Carton the cat

MAYBE
- [ ] Optimize retrieval
- [ ] Prevent from trying to download images too big

## :skull: This PR contains a major breaking change to the protocol

We will need to update all services and everyone will need to recreate their accounts, see my [comment](https://github.com/berty/berty/pull/2692#pullrequestreview-532614276) to review the offending code

The integration test is broken because of that I think


go/pkg/bertyprotocol/keystore_message.go
```diff
@@ -377,7 +377,13 @@ func (m *messageKeystore) OpenEnvelope(ctx context.Context, g *bertytypes.Group,
 		return nil, nil, errcode.TODO.Wrap(err)
 	}
 
-	return headers, msg, nil
+	var msg bertytypes.EncryptedMessage
+	err = msg.Unmarshal(msgBytes)
+	if err != nil {
+		return nil, nil, errcode.ErrDeserialization.Wrap(err)
```
n0izn0iz, line 383: :warning:  this is a major breaking change

zxxma:
👍 LGTM! 🔥

go/pkg/bertyaccount/service_grpc.go
```diff
@@ -247,3 +251,22 @@ func getSerivceError(err error) *Error {
 		Message:       err.Error(),
 	}
 }
+
+func (s *service) recv(id string, cstream *grpcutil.LazyStream) (*ClientStreamRecv_Reply, error) {
```
gfanton, line 255: can you update the name of this function to be more explicit?
(something like clientStreamRecv/streamRecv should be good)

go/pkg/bertyprotocol/api_attachment.go
```diff
@@ -0,0 +1,327 @@
+package bertyprotocol
```
aeddi, line 1: Could you keep in this file only the main API functions and move all the helpers to an `attachment_utils.go` file? (to be consistent with other api methods implem)

Could you also define most of your errors in errcode.proto or at least wrap them in an `errcode.TODO`?

go/pkg/bertyprotocol/api_attachment.go
```diff
@@ -0,0 +1,327 @@
...
+	defer plaintext.Close()
+
+	sk, ciphertext, err := attachmentSealer(plaintext, s.logger)
+	s.logger.Debug("stream cipher init done", zap.Error(err))
+	if err != nil {
+		return fmt.Errorf("failed to open attachment stream: %s", err.Error())
+	}
+	defer ciphertext.Close()
+
+	ipfsPath, err := s.ipfsCoreAPI.Unixfs().Add(stream.Context(), ipfsfiles.NewReaderFile(ciphertext))
```
aeddi, line 49: I'm not sure, is this enough to pin the content? (I didn't check IPFS doc)

go/pkg/bertyprotocol/keystore_device.go
```diff
@@ -243,6 +248,40 @@ func (a *deviceKeystore) RestoreAccountKeys(sk crypto.PrivKey, proofSK crypto.Pr
...
+	for _, at := range attachments {
+		cid, err := ipfscid.Cast(at.GetCID())
+		if err != nil {
+			return errcode.ErrDeserialization.Wrap(err)
+		}
+
+		s := at.GetSecret()
+		if len(s) <= len(attachmentKeyV0Prefix) || !bytes.Equal(s[:len(attachmentKeyV0Prefix)], []byte(attachmentKeyV0Prefix)) {
+			return errcode.ErrDeserialization.Wrap(fmt.Errorf("invalid secret prefix %s", s))
+		}
```
aeddi, line 269: I think this kind of check should be done in `api_attachment.go` or `attachment_utils.go` to centralize the logic of choosing the right alogrithm when multikey will be ready.
IMO this `PutAttachments` should just take two arrays of crypto.PrivKey and cIDs.

go/pkg/bertyprotocol/api_attachment.go
n0izn0iz: pin is the default but @Jorropo  says that it's not guaranteed in the future, will force it

go/pkg/bertyprotocol/keystore_device.go
n0izn0iz: agreed

api/bertyprotocol.proto
```diff
@@ -109,4 +109,10 @@ service ProtocolService {
 
   // PeerList returns a list of P2P peers
   rpc PeerList(types.v1.PeerList.Request) returns (types.v1.PeerList.Reply);
+
+  // PrepareAttachment ...
+  rpc PrepareAttachment(stream types.v1.PrepareAttachment.Request) returns (types.v1.PrepareAttachment.Reply);
+
+  // RetrieveAttachment returns an attachment data
+  rpc RetrieveAttachment(types.v1.RetrieveAttachment.Request) returns (stream types.v1.RetrieveAttachment.Reply);
```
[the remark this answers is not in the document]
n0izn0iz: It's a rule that makes code easier to write but harder to read which I think is not too good..

Anyway I'll stick to it :+1:

api/bertymessenger.proto
```diff
@@ -223,11 +223,13 @@ message AppMessage {
   message GroupInvitation {
     string link = 2; // TODO: optimize message size
   }
-  message SetGroupName {
-    string name = 2;
+  message SetGroupInfo {
+    string display_name = 1;
+    string avatar_cid = 2; // TODO: optimize message size
```
moul, line 228: I think we should add a `bool update_all` field or split the GroupInfo call for each available field:
* SetGroupName
* SetGroupAvatarCID

This is one of the rare "shared" entity, and because of the synchronicity of Berty in general, it's safer to allow someone to change just the title without requiring to be up-to-date for the other fields

To keep the API small, we can just keep `SetGroupInfo` and only set what is `!= ""`, in that case, we only need to make sure that the front code will only call SetGroupInfo with the "just-changed field" and not with all the group state known by the app

api/bertymessenger.proto
```diff
@@ -320,10 +323,12 @@ message Contact {
   Conversation conversation = 3;
   State state = 4;
   string display_name = 5;
+  string avatar_cid = 9  [(gogoproto.moretags) = "gorm:\"column:avatar_cid\"", (gogoproto.customname) = "AvatarCID"];
   int64 created_date = 7;
   // specific to outgoing requests
   int64 sent_date = 8;
   repeated Device devices = 6 [(gogoproto.moretags) = "gorm:\"foreignKey:MemberPublicKey\""];
+  int64 infoDate = 10;
```
moul, line 331:
```suggestion
  int64 info_date = 10;
```

api/bertymessenger.proto
```diff
@@ -320,10 +323,12 @@ message Contact {
   Conversation conversation = 3;
   State state = 4;
   string display_name = 5;
+  string avatar_cid = 9  [(gogoproto.moretags) = "gorm:\"column:avatar_cid\"", (gogoproto.customname) = "AvatarCID"];
```
moul, line 326:
```suggestion
  string avatar_cid = 9 [(gogoproto.customname) = "AvatarCID"];
```

sorry, I didn't check, but my current feeling is that the result is the same, nope?

api/bertymessenger.proto
```diff
@@ -207,8 +207,8 @@ message AppMessage {
     TypeUserMessage = 1;
     TypeUserReaction = 2;
     TypeGroupInvitation = 3;
-    TypeSetGroupName = 4;
-    TypeSetUserName = 5;
+    TypeSetGroupInfo = 4;
+    TypeSetUserInfo = 5;
```
moul, line 211: Ahah, I don't know if it was expected or if you were aware, but since you reuse the same number for those calls and since those calls still have the "name" as a string and number 1, we're backward-compatible on this part (if we re-read an old database, the marshaled protobuf message of `SetGroupName{Name: "foobar"}` will automatically be unmarshalled as `SetGroupInfo{DisplayName: "foobar"}` :)

api/bertymessenger.proto
```diff
@@ -377,7 +381,9 @@ message ConversationReplicationInfo {
 message Member { // Composite primary key
   string public_key = 1 [(gogoproto.moretags) = "gorm:\"primaryKey\""];
   string display_name = 2;
+  string avatar_cid = 6 [(gogoproto.moretags) = "gorm:\"column:avatar_cid\"", (gogoproto.customname) = "AvatarCID"];
   string conversation_public_key = 3 [(gogoproto.moretags) = "gorm:\"primaryKey\""];
+  int64 infoDate = 7;
```
moul, line 386:
```suggestion
  int64 info_date = 7;
```

api/bertytypes.proto
```diff
@@ -178,7 +181,8 @@ message GroupEnvelope {
   // event is encrypted using a symmetric key shared among group members
   bytes event = 2;
 
-  // TODO: Add more readable information here if necessary (eg. CIDs for replication service)
+  // encrypted_attachment_cids is a list of attachment CIDs encrypted specifically for replication services
+  repeated bytes encrypted_attachment_cids = 3;
```
moul, line 185:
```suggestion
  repeated bytes encrypted_attachment_cids = 3 [(gogoproto.customname) = "EncryptedAttachmentCIDs"];
```

api/bertytypes.proto
```diff
@@ -168,6 +168,9 @@ message GroupMetadata {
 
   // sig is the signature of the payload, it depends on the event_type for the used key
   bytes sig = 3;
+
+   // protocol_metadata is protocol layer data 
+   ProtocolMetadata protocol_metadata = 4;
```
moul, line 173:
```suggestion
  // protocol_metadata is protocol layer data 
  ProtocolMetadata protocol_metadata = 4;
```

api/bertytypes.proto
```diff
@@ -207,14 +225,15 @@ message MessageEnvelope {
   // nonce is a nonce for message headers
   bytes nonce = 3;
 
-  // TODO: Add more readable information here if necessary (eg. CIDs for replication service)
+  // encrypted_attachment_cids is a list of attachment CIDs encrypted specifically for replication services
+  repeated bytes encrypted_attachment_cids = 4;
```
moul, line 229:
```suggestion
  repeated bytes encrypted_attachment_cids = 4 [(gogoproto.customname) = "EncryptedAttachmentCIDs"];
```

api/bertytypes.proto
```diff
@@ -1143,3 +1171,36 @@ enum Direction {
...
+    bytes block = 1;
+
+    /* Header data:
+    ** IMPORTANT: the first request must contain the header data and an empty block, header data in following requests will be ignored
+    */
+  
+    // disable_encryption tells the protocol to store the file as plain text
+    bool disable_encryption = 2;
+
+
```
moul, line 1187:
```suggestion
```

api/bertymessenger.proto
n0izn0iz: >  only set what is != ""
this is what I'm doing (or tried at least ^^)

api/bertymessenger.proto
n0izn0iz: nope, it gets called "avatar_c_id" and breaks gorm internal stuff on specific queries

api/bertymessenger.proto
n0izn0iz: unexpected ^^

api/bertymessenger.proto
n0izn0iz: created issue upstream https://github.com/go-gorm/gorm/issues/3779

Makefile
```diff
@@ -4,6 +4,8 @@ all: generate test
 
 test:
 	cd go; $(MAKE) test
+	cd js; $(MAKE) test
+	cd js; $(MAKE) lint
 .PHONY: test
```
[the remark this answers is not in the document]
n0izn0iz: `cd go; $(MAKE) test` runs the go linter urrently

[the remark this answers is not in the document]
aeddi: Ok I didn't check, so keep it as it is :)

PR #2693, opened 2020-11-10
WIP: feat: allows to use libp2p for GRPC

Jorropo:


moul:
@Jorropo can you add examples of how to use this, please?

Is this simple to use while keeping a good level of security (i.e., auth, certs)?

Jorropo:
> Is this simple to use while keeping a good level of security (i.e., auth, certs)?

This is planned but not yet done (I tryed to fix the db stucking problem, ... And so right now I'm fixing bugs before adding new features)

aeddi:
Stale PR

PR #2695, opened 2020-11-10
fix(messenger): remove send on line return

clegirar:


clegirar:
I think it's better to remove it for the beta, this issue was frequently surveyed
And i can add this feature only on debug, because when simu was launch we earn time to just press enter to send a message, or just we can wait the desktop version because it's for this usecase that send message on press enter was usefull

moul:
isn't it a way to detect that this is a keyboard (on the simulator or like me on an iPad)?

btw, LGTM to merge this now and think later since we are helping the majority :+1:

clegirar:
I don't already dig on this
Okay 👍

Issue #2654, opened 2020-11-04
Improve Search screen when 'No Results found' 


zxxma:
**Current search status when "No Results" :** 

![current-noresults](https://user-images.githubusercontent.com/37345040/98128113-c7828400-1eb7-11eb-968d-a9b334bf731e.jpg)

**Expected**
![SearchMode-NoResults](https://user-images.githubusercontent.com/37345040/98128311-0e707980-1eb8-11eb-8865-1d64bb4bc6a7.png)

**Police of Quotes has to be :** 
Open Sans - Semi-Bold, and corresponding to this colors HEX:
Title 'Quote of the day' HEX: DADDE6
Quotes HEX: DBE1EC 

"No results found" should be in Yellow as in the search bar : HEX: FFAE3A

wamry:
I created a PR for this issue https://github.com/berty/berty/pull/2702

PR #2701, opened 2020-11-11
fix: No results found screen

wamry:
Fix No results found screen, related issue https://github.com/berty/berty/issues/2654


wamry:
work in progress

PR #2715, opened 2020-11-12
feat: export account data to tar file within mobile app

glouvigny:


moul (approved):
Tested, approved :+1:

PR #2718, opened 2020-11-12
chore: bump ipfs/libp2p deps

moul:
Closes #2708, #2710, #2711 #2722 
Depends on https://github.com/ipfs/go-ipfs/pull/7747


go/internal/handshake/handshake_test.go
```diff
@@ -25,7 +24,7 @@ func TestValidHandshake(t *testing.T) {
 		stream p2pnetwork.Stream,
 		mh *mockedHandshake,
 	) {
-		defer p2phelpers.FullClose(stream)
+		defer stream.Close()
```
Jorropo, line 27:
```suggestion
		defer stream.Reset()
```
Closes are async, it's more a free the buffers and stop saving what the other node is sending, I don't care it anyway.

To replace a `p2phelpers.FullClose` a `stream.Reset` is more suited.

moul: from what I understood, that's what changed in https://github.com/libp2p/go-libp2p-core/pull/166 and around

from what I understand `Close` is now the safe sync way to close a stream

can you double-check? I will adapt accordingly :+1:, thx

Jorropo: PS: Now that I think about that I have never checked the behaviour of `Reset`, I guess I should this later.

Jorropo: > 
> 
> would it be better to chain `Reset`, then `Close` in an helper?

Reset does Close : https://github.com/libp2p/go-yamux/blob/7a85a068c850f401ab7c0177ebafafc959e119bb/stream.go#L282 -> https://github.com/libp2p/go-yamux/blob/7a85a068c850f401ab7c0177ebafafc959e119bb/stream.go#L371
(Close = Close now whatever what and return)
(Reset = Append Close to buffer, wait for ack, then close and then return)

Stebalien: Note: `Reset` will _throw away_ any buffered data and signal an error. `Close` will make an effort to deliver the data, but will return before it finishes.

TL;DR: use close when you want to "finish off" a stream. Use `Reset` when you want to abruptly disconnect a stream.

If, instead, you want to _wait_ for any outstanding data to be received, you need to call `CloseWrite()` to flush and send an EOF, then read until you see an EOF from the other side, then finally call `Close()` to actually free resources.

Jorropo (approved):
I'm approving the part of @moul :)

go/internal/ipfsutil/helpers_test.go
```diff
@@ -0,0 +1,189 @@
...
+				err = s.Close() // Trying to close ourself
+				if err == nil {
+					errcb <- io.EOF // Perfect
+					return
+				}
+				errcb <- fmt.Errorf("error closing after EOF: %w", err)
+				return
+			}
+			if err != nil {
+				errcb <- fmt.Errorf("Error not EOF while reading: %w", err)
```
Jorropo, line 103:
```suggestion
				errcb <- fmt.Errorf("error not EOF while reading: %w", err)
```

Jorropo (changes requested):

go/internal/initutil/manager_test.go
```diff
@@ -31,13 +31,16 @@ func verifyRunningLeakDetection(t *testing.T) {
 	goleak.VerifyNone(t,
 		goleak.IgnoreTopFunction("github.com/desertbit/timer.timerRoutine"),                                                  // called by init() in github.com/desertbit/timer/timers.go, refer in grpc-web
 		goleak.IgnoreTopFunction("github.com/ipfs/go-log/writer.(*MirrorWriter).logRoutine"),                                 // global writer created at github.com/ipfs/go-log@v1.0.4/writer/option.go, refer by github.com/ipfs/, like go-bitswap
+		goleak.IgnoreTopFunction("github.com/jbenet/goprocess.(*process).doClose"),                                           // sometimes happening on CI, need more investigation
```
Jorropo, line 34: :'( many more

Jorropo (approved):
LGTM

go/pkg/bertyprotocol/contact_request_manager.go
```diff
@@ -316,7 +315,7 @@ func (c *contactRequestsManager) incomingHandler(stream network.Stream) {
 
 func (c *contactRequestsManager) performSend(otherPK crypto.PubKey, stream network.Stream) error {
 	defer func() {
-		if err := p2phelpers.FullClose(stream); err != nil {
+		if err := ipfsutil.FullClose(stream); err != nil {
```
Stebalien, line 318: This is only useful if you actually want to do something with the error (e.g., return it to indicate that the send didn't complete). This will wait a full round-trip.

If you just want to send, close, then walk away, just call `Close`.

go/pkg/bertyprotocol/contact_request_manager.go
```diff
@@ -267,7 +266,7 @@ const contactRequestV1 = "/berty/contact_req/1.0.0"
 
 func (c *contactRequestsManager) incomingHandler(stream network.Stream) {
 	defer func() {
-		if err := p2phelpers.FullClose(stream); err != nil {
```
Stebalien, line 270: I would definitely just call `stream.Close` here as there's nothing you can do if that fails.

PR #2738, opened 2020-11-18
fix(ci): downgrade go version in ios build

n0izn0iz:
This fixes the `go build runtime/cgo: invalid flag in go:cgo_ldflag: -fembed-bitcode` error


Jorropo (approved):
LGTM

PR #2749, opened 2020-11-19
fix(messenger): UI message monitor metadata

clegirar:


zxxma:
Maybe we should be more precise in Toggle title? Cause "Debug" is quite large.
"Debug infos in conversations" or something like that?  (imho)
but, great anyway!

LGTM 👍 🔥

clegirar:
Yes i agree 👍

js/packages/components/chat/message/Invitation.tsx
```diff
@@ -0,0 +1,216 @@
+import React, { useMemo, useState } from 'react'
+import { useStyles } from '@berty-tech/styles'
+import { Text as TextNative, TouchableOpacity, View } from 'react-native'
+import { Icon, Text } from '@ui-kitten/components'
+import { useClient, useConversation, useOneToOneContact } from '@berty-tech/store/hooks'
+import { Buffer } from 'buffer'
+import { ProceduralCircleAvatar } from '@berty-tech/components/shared-components'
+import { MessageSystemWrapper } from '@berty-tech/components/chat/message'
```
n0izn0iz, line 8:
```suggestion
import { ProceduralCircleAvatar } from '../../shared-components/ProceduralCircleAvatar'
import { MessageSystemWrapper } from './MessageSystemWrapper'
```
- please use relative imports for files in the same packages, I'm not sure the behavior around importing self is well defined: `import foo from "./bar"` instead of `import foo from "@org/pkg/bar"`
- please avoid recursive imports if you can `message/Invitation.tsx` -> `message` ->  `message/Invitation.tsx` (due to the `MessageSystemWrapper` import), the js ecosystem is smart around this but it can cause unexpected behaviors that are hard to debug
- please try to import only the minimum required: `import { MessageSystemWrapper } from "message/MessageSystemWrapper"` instead of `import { MessageSystemWrapper } from "message"`, this will help with unit testing

js/packages/components/chat/message/shared-components/User.tsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react'
```
n0izn0iz, line 1: I think we can flaten the tree by moving what's in "shared-components" to the parent folder

js/packages/components/chat/message/index.ts
```diff
@@ -0,0 +1,6 @@
+export { Message } from './Message'
```
n0izn0iz, line 1: I wouldn't expose all the message types here but only the polymorphic message component here, maybe even `mv Message.tsx index.ts`

if someone want only one message it can import the file of the message

n0izn0iz (commented):
this was needed, thanks

n0izn0iz (approved):
cursory lgtm

Issue #2614, opened 2020-10-30
Improve Swipe effects in app similar to Deck Swiper


zxxma:
We would love to improve swipes behaviors in app (like Deck Swiper), moving between desks.
Example: From Settings screen to Home screen (Conversation list), or in Profile, Scan QR, Create New group.

This lib can be used : 
https://github.com/alexbrillant/react-native-deck-swiper

The "moving screen" have also to be improved, with radius & blur on background.
Current behavior not optimized : 
![Capture_decran_2020-10-28_a_12 45 11](https://user-images.githubusercontent.com/37345040/97710160-ac7cd200-1abb-11eb-8ba9-8f6079cc4486.png)

sakul-budhathoki:
I am not sure of implementing Deck Swiper libraries with react-navigation. We can achieve swipe gesture; vertical or horizontal. And can customize animation and transition of the screen navigation. 
Here is example of horizontal swipe gesture and custom transition effect. 
![navigation](https://user-images.githubusercontent.com/13379067/99669933-dd876b80-2a97-11eb-814d-2ffe7f5eeabb.gif)

zxxma:
It looks neat @sakul-budhathoki 👍 
thanks for this proposal, let's go for this.

sakul-budhathoki:
Done in #2754

PR #2754, opened 2020-11-19
navigation transition

sakul-budhathoki:
Issues Link: #2614 

Changes: - 
- Replaced `react-native-screens/native-stack` by `@react-navigation/native`
- Applied custom transition 
- Removed ModalNavigation and used similar fade transition effect for modals
- Removed RequestSent component and it's navigation

navigation-custom


clegirar:
For the `RequestSent` screen, we used it in some last versions of Berty but the changes have meant that it's no longer used, so i can do a PR if you want for delete all of this ? Else you can delete screen type and the stack in the navigation, and the component
I have hesitated to keep it if in the short term future we would need it but i think we can delete it 👍

zxxma:
Nice job, looks neat.
LGTM 👍

n0izn0iz (approved):
code lvgtm :D

n0izn0iz:
@sakul-budhathoki please rebase and pass CI

PR #2756, opened 2020-11-19
feat: refactor sharing links

moul:
The goal of this PR is to refactor the sharing links system in order to make it more optimized and future-proof

## Current problems

* size can be optimized (useful for QR)
* not future-proof (i.e., this PR will break the past)
* https://berty.tech/id JS page does not support group and is difficult to keep synced
* some `go/cmd/` subcommand i.e. bots requires running a bertymessenger instance to just perform conversions

## Todo

* [x] create a super simple top-level API to manage everything related to links
* [x] ensure we have kind and versioning in the encoded protobuf
* [x] test various cases
* [x] fuzz test the parts with encoding
* [x] update https://berty.tech/id webpage to support new URLs (https://github.com/berty/www.berty.tech/pull/201)
* [x] update go code to use this new mini lib
* [x] rebase
* [x] ensure we will be able to support encryption
* [x] double-check missing FIXMEs
* [x] update the mobile code and entitlement files to support new URLs
* [x] add a check `[skip-ci]` or `[skip-integration]` in commits' message to actually do what it suggests
* [ ] testing session

---

Previous implementation: https://github.com/berty/berty/blob/f3184c7af55029362bc72ba85319456dd3224299/go/pkg/bertymessenger/api.go#L93-L119


go/internal/bertylinks/links.go
```diff
@@ -0,0 +1,45 @@
...
+		additionalFlags string // optional, contains human/JS readable variables, i.e., name
+	)
+
+	// Qr format
+	qr := "https://berty.tech/id#" + typeAndVersion + "/" + obscureBlob + "/" + additionalFlags
+
+	// contact sharing
+	{
+		typeAndVersion := 1 // contact-sharing version 1
+		obscureBlob := base62.Encode(proto.Marshal(BertyID{RDVP_SEED, ACCOUNT_PK}))
```
Jorropo, line 22: With padding or without ? That not clear :) (I guess without)

moul: What’s the best?

Jorropo: Depends, for all other keys around berty we used without.
It's the same except without if nothing if found the padding is implied.
So that make shorter keys so I would go for without.

moul: sounds good to me :+1:

Jorropo:
I think what you are doing is a bit like "this is the last final breaking change ! this one is for good the one" (I know it's not what you are telling and probably thinking but it feels a bit like it to me).

What if we discover that adding the peerid or tor maddr to the key makes the contact request 3 times faster ? How could I add this to the key and QR code without breaking stuff ?

**Why not doing like multiformat does ?**
## So my proposal is :

A single parser package **FONCTIONALIST, SIDE EFFECT FREE** would be used (very important later).

Binary structure, base building block :
```
<varuint type/id>[length for non static sized types]<blob>
```
Then the QR would be :
```
berty://<base62 unpadded blob>
```
Then the link would be :
```
berty.tech/id#<base62 unpadded blob>
```
Then when decoding we could depending of the reception context (berty.tech, qr red, ...) extract the base62 blob decode it and forward it to the `parse blob`.
Then `parse blob` would read the varuint (using multiformat code), switch case or a parser fonction lookup table the code.
Use the fonction found and pass the current blob as argument and return the output (`reponseStruct, error`).
What would be awsome with that is that we can have awesome self describing types and we can invent many thing to various needs easly.

If we are running out of space we can pack the blob in a `LZMA2` or `GZIP` block and the receiving end just call the same `parse blob` function, a compression function would just decompress his blob and return the result of `parse blob`.
And when a key id or something like that would be hit we would just build the `reponseStruct` object and start returning it.
We can even imagine some metadata types wich we could use to attach non essential metadata and code the metadata parser to ignore unkown metadata so if we want to add let's say the bootstrap addresses of a peer for anyreason the newer nodes would set some fields in the `responseStruct` but older would just ignore it.

The key type would be either a serializer we build ourself or some protobuf.

**Basicaly we would have a self describing recursing format.**
I think this solves all of our problems:
### size can be optimized (useful for QR)

Encoder can on demand deal with this :
```go
// Pseudo code
func CreateBlob(key keyStruct) (string, error) {
	blob, err := parsingutil.EncodeKey(key)
	if err != nil {
		return "", err
	}

	if len(blob) > maxSizeQr {
		compressionBlob := blob
		for compressionLevel := 0; len(compressionBlob) > maxSizeQr; compressionLevel++{
			if compressionLevel > 9 {
				return "", ErrKeyTooBig
			}
			compressionBlob = parsingutil.CompressLzma(blob, compressionLevel)
		}
		blob = compressionBlob
	}
	return "berty://" + encodeBase62(blob), nil
}
```

And the decoder would still be :
```go
// Pseudo code
key, err := parsingutil.Parse(blob)
checkErr(err)
switch key.Type {
case group:
	// ...
case oneToOne:
	// ...
}
```
Just like multiformat does. *Easy (for what this does) !*

### not future-proof (i.e., this PR will break the past)

I think we should try to avoid that also in the future (most of the crypto is set, we know how does we meet peers, how does we encrypt stuff, ... I we are in capabilities to say post that "all QR codes will forever be valid, at least in their core fonctionalities")

Let's assume I want to add a Lora frequency range in the key. But I don't want to break anything, ...
So I propose a special type to an option tag.
The optional tag have 4 thing :
```
<field id (varuint) or field name (string ?) (a field name is less colisiotion likely and more hyman readable if we choose to unsupervise the field creation process)><field length, varuint, (required by implementation who doesn't know this field to skip it)><field can be anything, this is set by the field id, might be a blob, float32><next data blob>
```
And this field would be special, implementation would not error on an unknown field, they would just skip it and proceed the parsing with the next data blob.

So if I want to add a Lora frequency range I could add it there, most of implementation would skip it but one doing Lora would read the frequency band and do some radio wizardry stuff with it.
What's cool with that is that this is infinitely expendable.

### berty.tech/id JS page does not support group and is difficult to keep synced

Best part yet :) (and not exclusive to my idea, it can be done with yours too)

If we stuck to some strict programing rules, making this package only returns simple types, no instanciation required to use it, few imports, ...
We can very easly use `go wasm` or `gopherjs` to generate some JS compatible code.
So in the JS we would just call the parser function and it could automatically dump all the key content, metadata, ...

### some go/cmd/ subcommand i.e. bots requires running a bertymessenger instance to just perform conversions

Same as the JS part.

moul:
> I think what you are doing is a bit like "this is the last final breaking change ! this one is for good the one" (I know it's not what you are telling and probably not thinking too but it feels a bit like it to me).

the goal of this PR is the exact opposite, make something that will support the future

> What if we discover that adding the peerid or tor maddr to the key makes the contact request 3 times faster ? How could I add this to the key and QR code without breaking stuff ?

then you add it in the protobuf part if it's only for machines or in the `url.Values{}` part if you need/want to make it easily parsable by human eyes or a JS script without wasm

> Why not doing like multiformat does ?

actually, that's exactly what we are doing with `proto.Marshal`, the only missing information is about the message version, which I suggested to use in clear in the URL to allow the frontend to easily detect it, but we can also add an enum in protobuf for the exact same purpose

those QR codes are only used by Berty Messenger (not even usable by Berty Protocol); it's completely ok to have an additional format later made for sharing with the rest of the p2p ecosystem, and it's not a problem to have both formats at the same time (just two output helpers)

switching to multiformat will add more constraint for us since we already use protobuf massively and we don't need to update any external repos to reserve some fields etc for equivalent advantages: binary, compression, versioning, deprecation, etc

> So my proposal is :

reminder: we need to have some clear parts readable by humans and usable by simple JS, so we need to have at least the type-version and custom tags available outside of the base62 encoded blob

switching to multiformats can only be considered for the specific machine-friendly part of the URL: `base62(proto.marshal(obj))`

I like your LZMA loop, we can reuse it with protobuf too

Jorropo:
> then you add it in the protobuf part if it's only for machines or in the `url.Values{}`

I see the point for human readability but that limits a lot the usefullness of compression. Crypto keys are high entropy data and doesn't compress very well while text compresses very easly, if space is really problematic wouldn't be possible to have a `map<string>string` in the protobuf, most of the time this field would be unset but if space need to be gained we can move metadata in this map where they would be compressed instead of in `url.Values{}`.

> JS script without wasm

`gopherjs` (I wasn't planning to use `wasm`, support is limited on old software) ? I know gopherjs isn't the most friendly transpiler for the programmers but for a parsing / encoding package it will not be too hard to force us to be gopher complient (it's mostly when you start doing networking, io, ...).

moul:
> I see the point for human readability but that limits a lot the usefullness of compression. Crypto keys are high entropy data and doesn't compress very well while text compresses very easly

Are you saying that it’s better to compress a text representation of a binary key that just use the binary key directly?

`len(gzip(base64(binkey)))` < `len(binkey)`?

I assume it’s totally possible but I prefer to see some proofs and statistics before adding complexity

> if space is really problematic wouldn't be possible to have a map<string>string in the protobuf, most of the time this field would be unset but if space need to be gained we can move metadata in this map where they would be compressed instead of in url.Values{}.

All the fields in the url.Values have their equivalent in the protobuf message, so it’s totally possible to make a short version of the QR where we move all the human values in the protobuf

I don't get how moving the clear `url.Values` into a `map` in the protobuf part could fit with the "we need a human readable part" requirement

(btw, a map in protobuf is the most unoptimized way to store data in term of size once marshaled)

> JS script without wasm; gopherjs (I wasn't planning to use wasm, support is limited on old browsers) ? I know gopherjs isn't the most friendly transpiler for the programmers but for a parsing / encoding package it will not be too hard to force us to be gopher complient (it's mostly when you start doing networking, io, ...).

Just out of curiosity, can you make a simulation of the size of the additional js bundle that contains all the deps to decode protobuf, lzma, base6X, etc? :)

Technically if we can manage everything in go, I also prefer a lot 👍 

In practice, I think that it's over-engineering that adds maintainability issues, that requires passing a lot of time to update the current website which is disconnected from the berty repo, generating a static website, having its own dev constraints etc

And btw, I don't see any benefits since we still need to keep a human readable part, which can be easily parsed using vanilla js if we use `url.Values{}` :)

Jorropo:
> Are you saying that it’s better to compress a text representation of a binary key that just use the binary key directly?

Not at all, I was trying to say that it's better to compress the username or `/onion3/aaaaaaaaaaaaaaaaaaaa...:1` than the key (interm of compression ratio, idealy we compress both).

moul:
So I gave various attempts tonight, here are my takeaways:

1. for the left encoded part which is basically: `proto.Marshal(struct{[32]byte, [32byte})`, compressing the binary results in most of the time (if not always) longer
2. for the same left encoded part, `compress(base64(leftPart))` is most of the time _(but not always)_, longer
3. the only compression algorithm that seems to have a better ratio/chance of being more efficient against our high-entropy binary could eventually be by using `zstd` with a dictionary (that should be kept with the codebase), more info here https://fastcompression.blogspot.com/2018/02/when-to-use-dictionary-compression.html
4. but even if we want small URLs, the main goal is to have short QR codes, and the good news is that there are various ways to make smaller QR codes, and one of the most efficient ones is to limit our alphabet to the "QR AlphaNum alphabet" which is a custom base45; I chose to specialize the web URL for readability and URL length and to optimize the internal URL for QR code scanning efficiency.
	4.1. 100 random tests with simple names https://gist.github.com/moul/7e63bcf300fff11f6aac56d457d6cf72
	4.1. 100 random tests with names containing Unicode https://gist.github.com/moul/d75447bc1b462f72411bd0ffeeb67f45

---

h/t @kimci86, thx for sharing his knowledge on compression and QR codes

Jorropo:
Might be a dumb question but my search results can't raise any answers.
QR codes specs store a raw binary blob.
We need `berty://`or `https://berty.tech/id` at the start of the QR code the the OS can open berty from the photo app.
But does the following data needs to be alphanumerical ? Our biggest space loss is clearly in the base62, if we could just do `berty://<binary compressed blob>` this would be way easier.

The only problem with that is that people who havn't installed berty wouldn't get redirected to the website.

moul:
two problems:
1. non-alphanum chars are escaped in URL representations (longer than baseXX)
2. QR code is more optimized when it can use its "alphanum alphabet" instead of switching to binary mode, and particularly when it needs to keep the `berty://` prefix

moul:
initial implementation is done and can be reviewed:

* go/internal/bertylinks/links.go
* go/internal/bertylinks/links_test.go

I plan to add more unit tests to support more malformed input cases

And then, I need to update the rest of the codebase to use the new library instead of the previous implementation

moul:
* I just finished to update all the go codebase to support the new links
* I can't find where we configure that the app is triggered for `berty://` deeplinks
* I will need some help to update the JS code (the .proto changed a little bit)
* I will then need some people to help me test various scenarios on all our clients (iOS app, mini CLI, berty.tech website)

PR #2774, opened 2020-11-20
fix(messenger): types for store providerEffect openingDaemon

clegirar:

### DIFF
--- a/js/packages/go-bridge/defaults.ts
+++ b/js/packages/go-bridge/defaults.ts
@@ -1,4 +1,6 @@
-export const GoBridgeDefaultOpts = {
+import { GoBridgeOpts } from './types'
+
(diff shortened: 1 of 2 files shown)


js/packages/go-bridge/defaults.ts
```diff
@@ -1,4 +1,9 @@
-export const GoBridgeDefaultOpts = {
+export type GoBridgeOpts = {
+	cliArgs: string[]
+	persistence: boolean
+}
```
n0izn0iz, line 4:
```suggestion
export type GoBridgeOpts = {
	cliArgs: string[]
	persistence: boolean
}
import { GoBridgeOpts } from "./types"
```

clegirar: Oh i don't see it ... sorry ... and thx 👍

Issue #2763, opened 2020-11-19
When a msg is very long in a conversation, we need to reduce it and and click on "Read more"


zxxma:
When people write very long messages, in 1to1 conversation and in Group conversations, we need a limit to prevent messages that are too long from taking up the user's entire screen.
Displaying a "Read more" at the end of the message and when user click, then deliver the all message.

In this example, it's in a group conversation.
For the moment, I propose to test a limit of **325 characters**.

![ReadMore](https://user-images.githubusercontent.com/37345040/99733162-0708bd00-2ac1-11eb-9d84-64297afd05b3.png)
.

sakul-budhathoki:
Done in #2776

PR #2776, opened 2020-11-22
read more in conversation

sakul-budhathoki:
Issue: #2763 

It will check for message length of 325 and show only 300 character along with ... and Read more.
Only showing 300 characters will support for message if length is for ex. 326, 327.

readmore


n0izn0iz (approved):
code lgtm
we might need to also limit to 20 '\n' max but it can be done later

PR #2792, opened 2020-11-23
fix: timeout while setuping

Jorropo:


go/internal/initutil/manager.go
```diff
@@ -123,6 +123,7 @@ type Manager struct {
 			listeners         []grpcutil.Listener
 		} `json:"GRPC,omitempty"`
 	} `json:"Node,omitempty"`
+	InitialisaitonTimeout time.Duration `json:"InitialisaitonTimeout,omitempty"`
```
moul, line 126:
```suggestion
	InitTimeout time.Duration `json:"InitTimeout,omitempty"`
```

(btw, it's `initialization`)

go/internal/initutil/manager.go
```diff
@@ -166,6 +167,11 @@ func New(ctx context.Context) (*Manager, error) {
 		m.Datastore.defaultDir = storageDirs[0].Path
 	}
 
+	// InitialisaitonTimeout
+	if m.InitialisaitonTimeout == 0 {
+		m.InitialisaitonTimeout = time.Minute
+	}
```
moul, line 173: nope, if timeout is 0 just disable timeout checking in prepareForGetter

Jorropo: > nope, if timeout is 0 just disable timeout checking in prepareForGetter

it is, that just for the default value, (I know that ~10 lines on top it says to put in CLI but I havn't made a CLI for that, do you think it's worth it ?)

go/internal/initutil/manager.go
```diff
@@ -188,8 +200,8 @@ func (m *Manager) GetContext() context.Context {
 
 func (m *Manager) RunWorkers() error {
 	m.workers.Add(func() error {
```
moul, line 202: missing a check for init timeout in mains that only call RunWorkers without getting anything first

go/internal/initutil/manager.go
```diff
@@ -267,3 +279,31 @@ func (m *Manager) AdvancedHelp() string {
...
+
+// prepareForGetter prepare locks and ctx timeouts for an external getter.
+// it returns a cleanup and should be defered.
+func (m *Manager) prepareForGetter() func() {
+	m.mutex.Lock()
+	if m.InitialisaitonTimeout >= 0 {
+		finishChan := make(chan struct{})
+		trueCtx := m.getContext()
+		var cancel func()
+		m.ctx, cancel = context.WithCancel(trueCtx)
```
moul, line 291: can you check if we are using m.ctx as initializer of some parts or just for init cancellation? 

I think that you may cancel things "after" the init, even if you restore the `trueCtx`

Jorropo: > I think that you may cancel things "after" the init, even if you restore the trueCtx

Who is you ?
- You is the goroutine lambda in `prepareForGetter`, it shouldn't because the `finishChan` is closed in the deffered returned lambda.
- You is the user who cancel his parrent context, it should because the ctx is pulled from the user with a `cancel`.

The only "dirty" thing is that the deffered lambda doesn't wait for the ticker watcher goroutine to be freed, I guess I can add a waitgroup but I don't think that important (at worst nothing happend and the goroutine spend some time doing nothing).

Jorropo: > can you check if we are using m.ctx as initializer of some parts or just for init cancellation?

Rereading makes me understand the question, it's used as ctx evrywhere as context but it shouldn't be canceled if the deffered lambda is executed.

go/internal/initutil/manager.go
moul: Yes, add a flag that sets it to 1 minute by default

go/internal/initutil/manager.go
moul: 👍

moul (approved):

PR #2793, opened 2020-11-23
chore: add a check to skip CI jobs

moul:
[skip-ci]

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,10 +1,6 @@
 name: Android
 on:
   push:
-    tags:
-      - v*
-    branches:
-      - master
     paths:
       - 'go/**'
       - '!go/**.md'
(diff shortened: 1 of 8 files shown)


.github/workflows/android.yml
```diff
@@ -25,6 +25,7 @@ on:
 jobs:
   build:
     name: Build for Android
+    if: "!contains(toJSON(github.event.commits.*.message), '[skip-ci]') && !contains(toJSON(github.event.commits.*.message), '[skip-android]' && !contains(toJSON(github.event.commits.*.message), '[skip-mobile]')"
```
n0izn0iz, line 28: won't that always skip on master once once commit with `[skip-ci]` is merged?

moul: not 100% sure, but I don't think so, those are not all the commits, but all the _event's_ commits

let's see :)

moul: wow, I'm comparing what other people are doing, one of the things that can become a GitHub action utility -> https://github.com/MinsukJi-NOAA/dummy-test/blob/cef36486a117aa0b1e9aea0f86f8d0ba6e0014cc/.github/workflows/manage.yml#L18-L32

PR #2797, opened 2020-11-23
WIP chore: add a way to skip ci checks with commit message

moul:
[skip-ci]


moul:
Can't get something cool with the current setup, current choices are:

One of the ways to achieve this is by making the .yml files a lot more complex: by adding preliminary steps + new conditions on each step (instead of just a condition at the job-level as I expect)

Check for the PR body instead of the commit messages

Wait that GitHub actions become more powerful (picked solution)

PR #2821, opened 2020-11-25
feat: bind go-libtor deps with iOS app

aeddi:
- [x] update `go-libp2p-tor-transport` to get the last version of `go-libtor`
- [x] download [static libs](https://github.com/berty/go-libtor/releases/tag/v1.0.333) required by `go-libtor` in Makefile (using `go.sum`)
- [x] build iOS gomobile framework with Tor embeded (linked with `go-libtor` static libs)
- [x] override `OpenSSL-Universal` used by RN with `go-libtor` openssl lib (see why [here](https://github.com/berty/go-libtor/issues/1))
- [x] link Berty app with `go-libtor` static libs

**Note:** Overriding `OpenSSL-Universal` pod required by RN using a local podspec could break the project at some point (e.g if we update RN and RN update the `OpenSSL-Universal` version, we will need _at least_ to update the version in our local podspec).
I tried to find a way to remove a pod transitive dependency (which could lead to similar issue anyway), but due to lack of time, I used this solution instead.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 require (
 	bazil.org/fuse v0.0.0-20200524192727-fb710f7dfd05 // indirect
 	berty.tech/go-ipfs-log v1.2.6
-	berty.tech/go-libp2p-tor-transport v0.8.2
+	berty.tech/go-libp2p-tor-transport v0.8.3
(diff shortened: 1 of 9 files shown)


gfanton (approved):
LGTM 👍

PR #2823, opened 2020-11-25
feat(messenger): add tor toggle and change daemon option

clegirar:
- [x] Add toggles in devtools that change tor async storage value
- [x] Check tor async storage value and add tor flag (`--tor.mode=required` or `--tor.mode=optional` or `--tor.mode=disabled`) when we run daemon


js/packages/store/context.ts
```diff
@@ -153,6 +154,10 @@ export type PersistentOptionsDebug = {
 	enable: boolean
 }
 
+export type PersistentOptionsTor = {
+	enable: boolean
```
Jorropo, line 158: Really not the good type.
accepted values are string :
- `disabled`
- `required`
- `optional` (an acceptable default if *unset* isn't suitable)
- *unset* (default)

Jorropo (changes requested):

js/packages/components/onboarding/SetupFinished.js
```diff
@@ -54,6 +54,12 @@ const SetupFinishedBody = () => {
 					enable: false,
 				},
 			})
+			await setPersistentOption({
+				type: PersistentOptionsKeys.Tor,
+				payload: {
+					flag: 'disabled',
```
Jorropo, line 60: Couldn't default to unset ?

js/packages/components/settings/DevTools.tsx
```diff
@@ -344,6 +347,40 @@ const BodyDevTools: React.FC<{}> = () => {
...
+				actionToggle={async () => {
+					await ctx.setPersistentOption({
+						type: PersistentOptionsKeys.Tor,
+						payload: {
+							flag: ctx.persistentOptions?.tor.flag === 'optional' ? 'disabled' : 'optional',
+						},
+					})
+				}}
+				toggled
+			/>
```
Jorropo, line 383: Couldn't be a dropdown menu instead ?

Jorropo (changes requested):

js/packages/components/settings/DevTools.tsx
clegirar: We decided that for the beginning to have 2 toggles for the beginning 👍

js/packages/components/onboarding/SetupFinished.js
clegirar: Yep i can 👍

Jorropo (approved):
SGTM (I can test it to LGTM it if you want)

PR #2825, opened 2020-11-25
fix: store state machine

clegirar:
Fix bridge flow for support restart option on screen stream error


glouvigny (approved):
LGTM

PR #2831, opened 2020-11-26
fix: Makefile rule for tor deps download

aeddi:
- [x] add `-mod=readonly` in `go list` command that retrieves `go-libtor` version
- [x] make download script exit on error


gfanton (approved):
LGTM 👍

PR #2837, opened 2020-11-27
feat(messenger): multimedia

n0izn0iz:
this also improves typings and fixes the group invitation in 1to1 and the network list


aeddi (approved):
LGTM 👍

PR #2870, opened 2020-12-07
feat: split bertymessenger and bertyprotocol packages

moul:
They will now have two parts each:

* the definitions: interfaces, enums, types (protobuf-based) + simple helpers  -> `pkg/protocoltypes` and `pkg/messengertypes`
* the implementation (importing the definitions) -> `pkg/bertyprotocol` and `pkg/bertymessenger`

It will help:

* to have a better godoc & smaller API surface
* to create `internal/` packages that manipulates protocol or messenger types while being imported by `bertyprotocol` and `bertymessenger` packages (avoid diamond deps issues)

---

todo

* [x] split packages (`api/*.proto`)
* [x] move bertylinks stuff in `internal/bertylinks`
* [x] port `go/` code
* [x] port `tool/` code
* [x] port `js/` code
* [x] port `docs/` code


js/packages/api/index.d.ts
```diff
@@ -5,7 +5,6 @@ import pb from './root.pb'
 import account = pb.berty.account.v1
 import messenger = pb.berty.messenger.v1
 import protocol = pb.berty.protocol.v1
-import types = pb.berty.types.v1
 import errcode = pb.berty.errcode
 
-export { account, messenger, protocol, types, errcode }
```
n0izn0iz, line 11: can you remove this pls ? what made this refactor easy was that I only allowed the root import "beapi" in my previous refactor

PR #2889, opened 2020-12-11
fix: pass log filters through gobridge

D4ryl00:
Logger filters must be given via grpc request to `openAccount` (https://github.com/berty/berty/blob/master/go/pkg/bertyaccount/service_account.go#L58)
The command line argument `--log.filters` is not used for mobiles.


moul (approved):

clegirar (commented):
LGTM 👍

PR #2910, opened 2020-12-17
WIP: DNS over TLS over Tor

Jorropo:
Depends on ncruces/go-dns#8 (optional)
Depends on berty/go-libp2p-tor-transport#34

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -77,7 +77,7 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.1 // indirect
 	github.com/mr-tron/base58 v1.2.0
 	github.com/multiformats/go-multiaddr v0.3.1
-	github.com/multiformats/go-multiaddr-dns v0.2.0
+	github.com/multiformats/go-multiaddr-dns v0.2.1-0.20201130213411-dba25a2c0b7a
 	github.com/multiformats/go-multiaddr-fmt v0.1.0
 	github.com/multiformats/go-multihash v0.0.14
 	github.com/multiformats/go-multistream v0.2.0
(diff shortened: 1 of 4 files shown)


ncruces:
Hi! I'd like to comment on some `go-dns` gotchas just so you're aware of them before proceeding. This seems more appropriate for this than the PR @ `go-dns`.

First one is that this doesn't work on Windows at all (I should make it explicit on the docs). This something Go needs to fix, there's little I can do about it.

Also, I see that you've decided to use DoT and not DoH. You should be aware that `go-dns` does not reuse DoT connections. A full TLS connection is setup and torn down for every DNS request. This might be seen as a privacy win, but it is probably especially slow over Tor.

Implementing connection sharing (correctly) for DoT is outside the scope of `go-dns`, because DoH gets it for free from the Go std library.

Finally, the cache implementation is intentionally very simple. It's a straight map with pseudo-random eviction. And I'd like to avoid taking on anything significantly more complicated. There may also be bugs in DNS message parsing, use at your own peril, but do submit reports and patches for those.

That said, I hope `go-dns` is useful and will work to merge that PR. Thanks!

Jorropo:
> First one is that this doesn't work on Windows at all (I should make it explicit on the docs). This something Go needs to fix, there's little I can do about it.

our code also breaks on windows,

> This might be seen as a privacy win

you got it right, if needed I'll add an option for timeout too (even if conns are reused this is still acceptable).

If performance is trully a problem I'll reconsider DoH, but dns is not critical in the application
only used rarely to resolve dns to content some times to times (also cached elsewhere) :
```console
$ dog -t TXT _dnslink.website.ipfs.io
TXT _dnslink.website.ipfs.io. 59s   "dnslink=/ipfs/bafybeiccywbfpsvdka53ozmi2r2mwf3j2mjwnhzjnfppjh2kyre5c6isqe"
```

aeddi:
Stale PR

aeddi:
Stale PR

PR #2916, opened 2020-12-17
fix(messenger): navigate after create group

clegirar:
Fix https://github.com/orgs/berty/projects/42#card-51408038


zxxma:
LGTM 👍

PR #2917, opened 2020-12-17
fix(messenger): ux edit profile

clegirar:
Fix https://github.com/orgs/berty/projects/42#card-51408684

* Before
Capture d’écran 2020-12-17 à 08 18 56
---
* With default avatar
Capture d’écran 2020-12-18 à 08 07 31

---
* With custom avatar
Capture d’écran 2020-12-18 à 08 07 42


zxxma:
I think we could just change the Pen (indicating the way to change avatar) with a Camera, 
it's a more common standard with in account editing pages.
I'll send you the .svg I used in the files menu!

zxxma:
Here is the reference for this screen : 
edit profile

zxxma:
LGTM 👍 
Just a detail, but could be interesting to upgrade : 
We should have the blur on the background, but not grey like on current version, it should display the Settings Screens in background like on the previous screen I uploaded in comment.

clegirar:
Update:
Capture d’écran 2020-12-21 à 12 21 13

zxxma:
LGTM 👍

Issue #1716, opened 2020-03-10
[chat-mvp] navigate to group on creation


n0izn0iz:
Currently after a group is created you have to navigate back manually.

We should probably open the group directly or navigate to the "Messages" screen

moul:
old, closing

PR #2926, opened 2020-12-21
feat: ui improvements

sakul-budhathoki:
- Audio Recording View
- Multiple files alignments fix
-  https://github.com/berty/bugs/issues/120
- https://github.com/berty/bugs/issues/100
- https://github.com/berty/bugs/issues/147

### DIFF
--- a/js/packages/components/chat/MultiMember.tsx
+++ b/js/packages/components/chat/MultiMember.tsx
@@ -49,7 +49,7 @@ const HeaderMultiMember: React.FC<{
 	stickyDate?: number
 	showStickyDate?: boolean
 }> = ({ id, stickyDate, showStickyDate }) => {
-	const { navigate } = useNavigation()
+	const { navigate, goBack } = useNavigation()
 	const [{ row, padding, flex, text, column, margin, color }] = useStyles()
 	const conversation = useConversation(id)
 	const [layoutHeader, onLayoutHeader] = useLayout() // to position date under blur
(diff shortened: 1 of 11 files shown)


zxxma:
Looks good to me! :) thanks Sakul!

js/packages/components/chat/MultiMember.tsx
```diff
@@ -73,7 +73,7 @@ const HeaderMultiMember: React.FC<{
 			>
 				<TouchableOpacity
 					style={[flex.tiny, flex.justify.center, flex.align.center]}
-					onPress={goBack}
+					onPress={() => navigate.main.home()}
```
n0izn0iz, line 76: are you sure these changes are wanted ?

n0izn0iz (commented):
lgtm except question

js/packages/components/chat/MultiMember.tsx
sakul-budhathoki: Hi @n0izn0iz , I have updated this code, as we need to navigate to home screen from Group Chat when new group is created it redirect to Group Chat. If goBack is implemented then it will go back to Group Creation screen.

n0izn0iz: The right fix ix to use  `reset` instead of `navigate` in Group Creation then

n0izn0iz: I believe this is addressed in #1716

n0izn0iz: I believe the fix is already merged in https://github.com/berty/berty/pull/2916

PR #2936, opened 2020-12-22
fix: state machine to support new multi-account UI/UX

clegirar:
- Refacto `js/packages/components/main/Home.tsx`
- Change the front state machine for support the future multi-account UI/UX
- Add the UI base of the multi account
- Add the UI base of screen stream gate (screen that show steps of the current stream)

I'll create an other PR for optimize UI of both screens

⚠️ Import account feature doesn't work (with the new UI, the button doesn't exist, but we'll thinking about it, but the method in the store was depecrated) ⚠️


n0izn0iz (approved):
LGTM

n0izn0iz (changes requested):
bugs detected during testing

zxxma:
LGTM (front side ok) 👍

PR #2997, opened 2021-01-11
wip: Better DB reads handling

Jorropo:


go/pkg/bertymessenger/db.go
```diff
@@ -470,10 +471,15 @@ func (d *dbWrapper) addContactRequestOutgoingSent(contactPK string) (*messengert
 		}); res.Error != nil {
 		return nil, res.Error
 	} else if res.RowsAffected == 0 {
-		return nil, errcode.ErrDBAddContactRequestOutgoingSent.Wrap(fmt.Errorf("nothing found"))
+		// Check if that not already in the DB.
+		res := d.db.Where(&messengertypes.Contact{PublicKey: contactPK}).First(&contact)
```
moul, line 475: hum, I think we need someone else to check if this not a problem, because we are explicitly taking a contact with any State

can someone else give a look please?

go/pkg/bertymessenger/handlers.go
```diff
@@ -295,10 +295,11 @@ func (h *eventHandler) accountContactRequestOutgoingEnqueued(gme *protocoltypes.
 	contactPKBytes := ev.GetContact().GetPK()
 	contactPK := b64EncodeBytes(contactPKBytes)
 
+	// FIXME: Discuss about removing this whole Metadata unmarshaling by setting `ShareableContact.metadata` type directly to `ContactMetadata`.
 	var cm messengertypes.ContactMetadata
-	err := proto.Unmarshal(ev.GetContact().GetMetadata(), &cm)
-	if err != nil {
-		return err
+	if err := proto.Unmarshal(ev.GetContact().GetMetadata(), &cm); err != nil {
```
moul, line 300: did this happen while you were testing?

it's maybe not "critical", it's really strange if it happens and I'm not sure it's a good thing to ignore it at this stage -> need additional reviews

go/pkg/bertymessenger/service.go
```diff
@@ -309,6 +310,7 @@ func (svc *service) subscribeToMetadata(gpkb []byte) error {
 	if err != nil {
 		return errcode.ErrEventListMetadata.Wrap(err)
 	}
+	runtime.Breakpoint()
```
moul, line 313:
```suggestion
```

go/pkg/bertymessenger/handlers.go
Jorropo: Yes this does happend, this is mini using a pure string instead of a protobuf struct.

aeddi:
Stale PR

PR #3021, opened 2021-01-21
chore: cleanup assets and iOS project

### LINKED ISSUE #2835: Rearrange common assets
Arrange the app's common assets neatly in the `js/assets` folder with sub-folders named by type:
- fonts
- audio (e.g notification sounds)
- images (e.g default avatar)
- etc...

And properly wire the different layers of the app (js/RN, iOS and Android) so that they can access them. For example: add a "Build Phase" in XCode that copies the sound files of the notifications to the IPA.

**Note**: Android or iOS specific assets can stay in their respective folders (e.g. app icons, launchscreens, etc...).

aeddi:
Fixes #2835


moul (approved):

js/ios/berty.yaml
```diff
@@ -138,8 +138,11 @@ targets:
     - path: ./Configs/GoogleService
       group: Configs
       buildPhase: none # will be added by a script
-    - path: ../assets
-      name: OpenSans
+    - path: ../assets/font
```
n0izn0iz, line 141: we need to adjust the makefile deps too

n0izn0iz (changes requested):
lgtm except comment

js/ios/berty.yaml
aeddi: Yes you're right, I'll add this deps

n0izn0iz (approved):
`js/assets` feels duplicate with `js/packages/assets`
I'm for putting stuff in `js/packages/assets` so we can import them in js code with `@berty-tech/assets` and keep the same structure as for the other js importables

aeddi:
> `js/assets` feels duplicate with `js/packages/assets`
> I'm for putting stuff in `js/packages/assets` so we can import them in js code with `@berty-tech/assets` and keep the same structure as for the other js importables

Okay, as long as it's cleaner, I'm fine with it 👍 
You'll do this in another PR if I understand correctly, I'll merge this one.

PR #3024, opened 2021-01-22
feat: add gomobile ipfs

gfanton:
Add gomobile ipfs

depends on https://github.com/ipfs-shipyard/gomobile-ipfs/pull/61


go/Makefile
```diff
@@ -166,6 +166,7 @@ $(gen_sum): $(gen_src)
 	  docker run \
 	    --user="$$uid" \
 	    --volume="`go env GOPATH`/pkg/mod:/go/pkg/mod" \
+		--volume="$(GOPATH)/src/github.com/ipfs-shipyard/gomobile-ipfs:/go/src/github.com/ipfs-shipyard/gomobile-ipfs" \
```
moul, line 169: I think that you should use the `go list` command here (as we've already done somewhere else in the makefiles)

moul (approved):

go/Makefile
gfanton: this was for testing purpose, since I was using local `gomobile-ipfs` repo, i will remove this change, I dont need it anymore !

aeddi (approved):
LGTM 👍

Issue #3041, opened 2021-01-28
bertyaccount's blackbox test fails when run quickly


n0izn0iz:
on current master:

```
--- FAIL: TestFlow (0.98s)
    blackbox_test.go:268: 
        	Error Trace:	blackbox_test.go:268
        	Error:      	Should be true
        	Test:       	TestFlow
FAIL
FAIL	berty.tech/berty/v2/go/pkg/bertyaccount	1.961s
```

line 268 is:
```golang
require.True(t, lastProgress.Delay > uint64(time.Duration(10*time.Millisecond).Microseconds())) // in general, it's around 40ms on Manfred's Linux server
```

moul:
I've other issues i.e., a close that does not wait enough time for all actions to be executed, but I can't reproduce the failing test, can you recheck @n0izn0iz?

![Screenshot - Manfred 2021-02-08 at 08 50 15](https://user-images.githubusercontent.com/94029/107190103-ae273180-69ea-11eb-998e-83a89478d9c9.png)

PR #3042, opened 2021-01-28
test: add a speedy test suite

Jorropo:
I'm surprised my self that we need it but somehow our current master (5febf9ca969d400925cb32b91b15fcee899df3c7) test suite is working only because we have `-race` enabled.
This will hopefully let us catch this kind of bug next time.


n0izn0iz:
related to #3041

Jorropo:
Sadly the CI is too slow and the test pass even if it doesn't on our machines.

PR #3060, opened 2021-01-31
chore: update connect HTML template

cdeleeuwe:


moul:
@cdeleeuwe can you try a version that does not make any external requests for the assets?

if really not possible to embed everything in the HTML, try to put all the assets locally and we will help generate an embedded page

cdeleeuwe:
> @cdeleeuwe can you try a version that does not make any external requests for the assets?
> 
> if really not possible to embed everything in the HTML, try to put all the assets locally and we will help generate an embedded page

@moul sure, I removed the external font and uses a local font instead.
image

moul (approved):

PR #3099, opened 2021-02-15
fix: play sound notifications depending on notif setting

clegirar:


n0izn0iz (changes requested):
I would prefer that we extend `playSound` in the context and bind the option there
If not, please keep the previous interface and use if statement at call sites

n0izn0iz (approved):
LGTM

zxxma:
Still ringing on real devices (silence mode: on) :
- on Beta Bot notifications at launch
- on every msg notifications

zxxma:
LGTM 👍  fixed 🔉  & 🔇  :)

PR #3100, opened 2021-02-15
feat: new conversation footer

clegirar:
Signed-off-by: clegirar <clemntgirard@gmail.com>

PR #3107, opened 2021-02-17
feat: hold/release to record voice message

glouvigny:


glouvigny:
Integrated into #3100

PR #3138, opened 2021-03-08
feat: exposing a message text search method from go

### LINKED ISSUE #2083: Messenger search does not work with very large volumes of data
Results need to be retrieved and displayed incrementally.

In the meantime, we can at least set some cutoff limit for number of items to search.

At the moment, if we have a million messages and search the letter 'a', search feature will not be too usable. 

Consider using fusejs (already installed it seems) and/or different implementations of delivering search results so we can achieve this.

(Fusejs or similar could also help with fuzzy searches, etc.)

glouvigny:
fixes #2083


clegirar (approved):
LGTM 👍 (js side)

aeddi (approved):
LGTM 👍

PR #3149, opened 2021-03-11
fix: network config navigation

clegirar:


zxxma:
I think the behavior for netfwork config could be optimize like this : 
- when we arrive in the home, the user click on the "message" inviting to Network configuration.
and then, the pop up configuration open.

Because currently, we don't have time to see messages for cfg in the list, the pop up is directly open.
Apart from that, it's perfect!

zxxma:
LGTM 👍

PR #3172, opened 2021-03-18
fix: ui sticky player

clegirar:


zxxma:
Some corrections needed : 
- when we're in a conversation, if we press play with the conversation still open, we don't need to open the sticky player. It should only open the sticky if we move out of this conversation. If we need to play/pause, we already got the main file in the conversation.
- the 🔉 button (top right) should mute sound when we click on.
Apart from that, it's great 👍

PR #3188, opened 2021-03-22
feat(messenger): replies and reactions backend

clegirar:


go/pkg/bertymessenger/service.go
```diff
@@ -526,6 +537,59 @@ func (svc *service) sendAccountUserInfo(groupPK string) error {
...
+		); err != nil {
+			return errcode.ErrMessengerStreamEvent.Wrap(err)
+		}
+	}
+	return nil
+}
+
+func buildReactionsView(tx *dbWrapper, cid string) ([]*mt.Interaction_ReactionView, error) {
+	reactions := ([]*mt.Reaction)(nil)
+	if err := tx.db.Where(&mt.Reaction{TargetCID: cid, State: true}).Find(&reactions).Error; err != nil {
```
glouvigny, line 565: You might want to replace this query by a raw sql query which would be more efficient, it would look something like:

```sql
SELECT count(*) AS total_reactions, emoji, MAX(is_mine) > 0 AS has_mine
FROM reaction
WHERE target_cid = "QmA"
GROUP BY emoji
ORDER BY MIN(state_date) ASC```

https://www.db-fiddle.com/f/kLHpiAo5XQMXb7J12LGsUo/0

File: api/messengertypes.proto [hunk 2]
```diff
@@ -417,6 +407,14 @@ message Interaction {
   bool acknowledged = 10;
   string target_cid = 13 [(gogoproto.moretags) = "gorm:\"index;column:target_cid\"", (gogoproto.customname) = "TargetCID"];
   repeated Media medias = 15;
+
+  // specific to client model
+  repeated ReactionView reactions = 16 [(gogoproto.moretags) = "gorm:\"-\""];
```
glouvigny, line 412: Name should be changed imo, view is used in too many contexts already

api/messengertypes.proto
```diff
@@ -810,3 +809,12 @@ message AudioPreview {
...
 }
+
+message Reaction {
+  string target_cid = 2 [(gogoproto.moretags) = "gorm:\"column:target_cid;primaryKey\"", (gogoproto.customname) = "TargetCID"];
+  string member_public_key = 3 [(gogoproto.moretags) = "gorm:\"primaryKey\""];
+  string emoji = 4 [(gogoproto.moretags) = "gorm:\"primaryKey\""];
+  bool is_mine = 5;
+  bool state = 6;
+  int64 state_date = 7;
+}
```
glouvigny, line 820: I think you has a problem with composite primary keys in an earlier PR. Maybe you can just use a cid as the pk and wrap write on this table with transactions to remove previous data.

Can we remove the state field and delete entries if they are no longer required instead?
Can we use sent_date for consistency with other models?
Can we remove is_mine from models and handle this with a value provided globally?

n0izn0iz:
opening a new PR cause github is preventing me from pushing on this one for some reason
(hence I can't rebase and move it forward)

api/messengertypes.proto
n0izn0iz: open to proposals ^^

api/messengertypes.proto
n0izn0iz: - state_date is specific to the last time the state was set so I prefer this way
- I agree but it will be a tricky query and and optimization so we can do it later

go/pkg/bertymessenger/service.go
n0izn0iz: thanks

PR #3197, opened 2021-03-25
feat: reply in chat

sakul-budhathoki:

---
name: New pull request
about: Create a new pull request
title: "feat:  Reply in chat"
---


zxxma:
Looks nice!
I think we have to optimize something : 
When here I replied to the bot with a "Hello", we should only have a preview of the previous message I replied, and not the full message. The idea is to have a character limit corresponding to the width of the message, so that it gives "the first sentence" then "...".
Then when the user clicks on the preview, he is automatically returned to the original message. 
And not like on this screen, where the "old message" I'm replying to, is full displayed.

**Current test on this PR**
screenreply

**Example of what's expected, with a shorter version of the previous msg finishing by '...' after on line maximum**
_(in this screen, the message is not sent yet, but it's the same way it's expected)_

Replying

n0izn0iz (approved):
LGTM

zxxma:
Looks great, some little bugs we should fix before merging : 
- Reactions have to be aligned on right of the msg
- We would like to work on a new Reply UX when the user swipe right.
We think that this lib could fit with the need : https://github.com/jemise111/react-native-swipe-list-view
Could you check it?

clegirar (approved):
LGTM 👍 (code + test iOS/android)

zxxma:
LGTM 👍 (Android + iOS)

Just a note : 
When we click on a replied message, we should come back to the initial message.
But it's ok to merge and can be done later!

PR #3218, opened 2021-04-03
fix(messenger): smooth navigation

clegirar:


zxxma:
- QR not accessible in the Profile could be a problem, adding a step in the sharing QR process
We keep it in WIP for the moment.

clegirar:
This PR brings us too little fluidity (or not at all) and remove all screen transitions, so we decided to close this PR and maybe re-open if the navigation was too slow 👍

PR #3266, opened 2021-04-20
feat: pressable avatar images

sakul-budhathoki:


n0izn0iz (approved):
LGTM (code review)

zxxma:
Looks like it doesn't work, or maybe not understood "where" is pressable?
(not working in a Group conversation)

sakul-budhathoki:
Hi @zxxma . I understand the problem, I will update it and let you know.

clegirar (approved):
Code LGTM 👍

zxxma:
Problem detected, code looks ok, but it's still not clickable in the app:
image

Same in the Group Settings or in Conversation, when we press, it doesn't work.

sakul-budhathoki:
For now it will be clickable if it is an image, above shown avatar is actually a text inside a circle. I will create a separate view to show this kind of avatar.

zxxma:
You're right, my bad!
LGTM 👍

PR #3267, opened 2021-04-20
feat(tyber): add lot of traces, network listen, events subscribe, step mutators, trace rename/reopen, codegen and websocket server

n0izn0iz:


tool/tyber/go/go.mod
```diff
@@ -1,8 +1,9 @@
-module berty.tech/berty/tool/tyber/go/v2
+module berty.tech/berty/tool/tyber/go
 
 go 1.13
 
 require (
+	berty.tech/berty/v2 v2.273.5 // indirect
```
moul, line 6:
```suggestion
	berty.tech/berty/v2 latest // indirect
```

tool/tyber/go/go.mod
```diff
@@ -1,8 +1,9 @@
-module berty.tech/berty/tool/tyber/go/v2
```
moul, line 1: hum, I don't understand why you're doing this here?

n0izn0iz: it was breaking the linux build with `/v2`

go/pkg/bertyaccount/service_account.go
```diff
@@ -171,10 +172,13 @@ func (s *service) openAccount(req *OpenAccount_Request, prog *progress.Progress)
 }
 
 // OpenAccount starts a Berty node.
-func (s *service) OpenAccount(_ context.Context, req *OpenAccount_Request) (*OpenAccount_Reply, error) {
+func (s *service) OpenAccount(ctx context.Context, req *OpenAccount_Request) (_ *OpenAccount_Reply, err error) {
 	s.muService.Lock()
 	defer s.muService.Unlock()
 
+	_, _, endSection := tyber.Section(ctx, s.logger, fmt.Sprintf("Opening account %s (AccountService)", req.AccountID))
+	defer func() { endSection(err, "") }()
```
moul, line 180: Can you double check that if you put an `err = fmt.Errorf()` below, it will be ported here too, I've a doubt that you need to use a `var err *error` manually

tool/tyber/go/go.mod
moul: are you using astilectron latest version? because it's similar with gomobile when gomobile was 1 or 2 versions below go master and not supporting go modules well at that time

tool/tyber/go/parser/header.go
```diff
@@ -29,14 +31,19 @@ type Header struct {
...
 			Protocol struct {
-				SwarmListeners        string        `json:"swarmListeners,omitempty"`
-				IPFSAPIListeners      string        `json:"iPFSAPIListeners,omitempty"`
-				IPFSWebUIListener     string        `json:"iPFSWebUIListener,omitempty"`
-				Announce              string        `json:"announce,omitempty"`
-				NoAnnounce            string        `json:"noAnnounce,omitempty"`
-				LocalDiscovery        bool          `json:"localDiscovery,omitempty"`
-				Ble                   bool          `json:"ble,omitempty"`
-				MultipeerConnectivity bool          `json:"multipeerConnectivity,omitempty"`
+				SwarmListeners    string `json:"swarmListeners,omitempty"`
```
moul, line 34: `swarmListeners` -> starting with lowercase
`Format` -> starting with uppercase

this inconsistency is very strange

go/framework/bertybridge/bridge.go
```diff
@@ -65,15 +77,63 @@ func NewBridge(config *Config) (*Bridge, error) {
...
-	{
-		if nativeLogger := config.dLogger; nativeLogger != nil {
-			b.logger = newLogger(nativeLogger)
-		} else {
-			b.logger = zap.NewNop()
+
+	cores := []zapcore.Core(nil)
+
+	// tyber core
+	if len(config.TyberHost) != 0 {
```
moul, line 84:
```suggestion
	if config.TyberHost != "" {
```

go/framework/bertybridge/bridge.go
```diff
@@ -65,15 +77,63 @@ func NewBridge(config *Config) (*Bridge, error) {
...
+	// create logger
+	if len(cores) == 0 {
+		b.logger = zap.NewNop()
+	} else {
+		b.logger = zap.New(
+			zapcore.NewTee(cores...),
+			zap.AddCaller(),
+		)
+		b.logger.Debug("logger initialized", zap.Any("manager", &initutil.Manager{
+			SessionID: randomID(),
```
moul, line 125: this is super hacky, you should just let initutil.Manager makes the stuff by itself

what should be done:
* update `bertyaccount/service_account.go` to use `initutil.New()` instead of `initutil.Manager{}`, so it creates an uuid session.ID
* call `manager.SetLogger(b.logger)`

and at the end, you can do this
```suggestion
		b.logger.Debug("logger initialized", zap.Any("manager", manager))
```

go/pkg/bertyaccount/service.go
```diff
@@ -96,7 +100,10 @@ func NewService(opts *Options) (Service, error) {
 	return s, nil
 }
 
-func (s *service) Close() error {
+func (s *service) Close() (err error) {
+	_, _, endSection := tyber.Section(context.TODO(), s.logger, "Closing AccountService")
```
moul, line 104: shouldn't this use something like `s.ctx` instead of `context.TODO()`?

go/pkg/bertyaccount/service_account.go
```diff
@@ -393,10 +406,13 @@ func (s *service) getAccountMetaForName(accountID string) (*AccountMetadata, err
...
-func (s *service) DeleteAccount(_ context.Context, request *DeleteAccount_Request) (*DeleteAccount_Reply, error) {
+func (s *service) DeleteAccount(ctx context.Context, request *DeleteAccount_Request) (_ *DeleteAccount_Reply, err error) {
 	s.muService.Lock()
 	defer s.muService.Unlock()
 
+	_, _, endSection := tyber.Section(ctx, s.logger, fmt.Sprintf("Deleting account %s (AccountService)", request.AccountID))
+	defer func() { endSection(err, "") }()
+
 	if s.initManager != nil {
 		return nil, errcode.ErrBertyAccountAlreadyOpened
```
moul, line 417: is the err in `endSection` set to `errcode.ErrBertyAccountALreadyOpened` here? or is it just returned like this with the `endSection(nil, "")`

go/pkg/bertyaccount/service_account.go
```diff
@@ -393,10 +406,13 @@ func (s *service) getAccountMetaForName(accountID string) (*AccountMetadata, err
 	return meta, nil
 }
 
-func (s *service) DeleteAccount(_ context.Context, request *DeleteAccount_Request) (*DeleteAccount_Reply, error) {
+func (s *service) DeleteAccount(ctx context.Context, request *DeleteAccount_Request) (_ *DeleteAccount_Reply, err error) {
 	s.muService.Lock()
 	defer s.muService.Unlock()
 
+	_, _, endSection := tyber.Section(ctx, s.logger, fmt.Sprintf("Deleting account %s (AccountService)", request.AccountID))
+	defer func() { endSection(err, "") }()
```
moul, line 414: maybe you sometimes use the two first anonymous variables, but since I only see this usage, I suggest to add a dedicated helper for this most common way of using the app to increase the readability, something like:

```suggestion
	defer tyber.DeferableSection(ctx, s.logger, fmt.Sprintf("Deleting account %s (AccountService)", request.AccountID))(err, "")
```

or even

```suggestion
	defer tyber.DeferableSection(ctx, s.logger,
		fmt.Sprintf("Deleting account %s (AccountService)", request.AccountID),
	)(err, "")
```

go/framework/bertybridge/bridge.go
```diff
@@ -2,24 +2,29 @@ package bertybridge
```
moul, line 2: @gfanton can you double-check this file please?

go/framework/bertybridge/bridge.go
moul: if my PR is merged, I suggest you to just rebase master and apply the last changes -> https://github.com/berty/berty/pull/3334

go/pkg/bertyaccount/service_account.go
n0izn0iz: no, see this playground https://play.golang.org/p/ejXulLs39tt

go/pkg/bertyaccount/service_account.go
n0izn0iz: as you can see in the playground https://play.golang.org/p/ejXulLs39tt if you don't use a lambda, the error is bound at the start and so is nil

n0izn0iz: I will add an helper without the two first returns

go/pkg/bertyaccount/service.go
n0izn0iz: Well I could in this case but it's tricky
The service might have a traceID in the context and so it won't create a new trace
I wanted to be sure to create a new trace, I should add an helper for that instead of using context.TODO

go/framework/bertybridge/bridge.go
n0izn0iz: this happens before the manager gets initialized
I think it's good that the logger gets initialized first so we get nformation even in case of early errors
I think it would be better to define a format for the tyber header instead of logging the manager struct

go/framework/bertybridge/bridge.go
```diff
@@ -226,8 +286,14 @@ func (b *Bridge) WillTerminate() {
 	}
 }
 
-func (b *Bridge) Close() error {
-	b.logger.Info("Bridge.Close called")
+func (b *Bridge) Close() (err error) {
+	_, _, endSection := tyber.Section(context.TODO(), b.logger, "Closing Berty framework")
+	defer func() {
```
gfanton, line 291: no need to **defer** this

go/framework/bertybridge/bridge.go
gfanton: > I think it's good that the logger gets initialized first so we get nformation even in case of early errors

No, it's clearly not worth it, early error should simply prevent the app to start, it's far more confusing to have this here, especially without any comment around 

> I think it would be better to define a format for the tyber header instead of logging the manager struct

we should definitely avoid the use of `Any` and only log useful field of the manager (if you really want to log this)

go/internal/logutil/tyber.go
```diff
@@ -0,0 +1,39 @@
...
+	if !strings.Contains(host, ":") {
+		host += ":4242"
+	}
+	logger := TyberLogger{host: host}
+	var err error
+	logger.conn, err = net.Dial("tcp", host)
+	if err != nil {
+		return nil, errcode.TODO.Wrap(err)
+	}
+	return &logger, nil
```
gfanton, line 29: use a context here
```suggestion
func NewTyberLogger(ctx context.Context, host string) (*TyberLogger, error) {
	host = strings.TrimSpace(host)
	if !strings.Contains(host, ":") {
		host += ":4242"
	}

	var d net.Dialer
	conn, err := d.DialContext(ctx, "tcp", host)
	if err != nil {
		return nil, errcode.TODO.Wrap(err)
	}

	return &TyberLogger{
		conn: conn,
		host: host,
	}, nil
}

```

go/pkg/bertyaccount/service_account.go
gfanton: `err` is never set since `err := s.openAccount(req, nil)` use `:=` instead of `=` bellow

go/pkg/bertyaccount/service.go
```diff
@@ -72,7 +73,10 @@ func (o *Options) applyDefault() {
 	}
 }
 
-func NewService(opts *Options) (Service, error) {
+func NewService(opts *Options) (_ Service, err error) {
+	_, _, endSection := tyber.Section(context.TODO(), opts.Logger, "Initializing AccountService")
+	defer func() { endSection(err, "") }()
```
gfanton, line 78: why don't you simply use `defer endSection(err, "")`?

n0izn0iz: see https://play.golang.org/p/ejXulLs39tt

aeddi (changes requested):

go/pkg/bertyprotocol/api_contact.go
```diff
@@ -7,9 +7,13 @@ import (
 
 	"berty.tech/berty/v2/go/pkg/errcode"
 	"berty.tech/berty/v2/go/pkg/protocoltypes"
+	"berty.tech/berty/v2/go/pkg/tyber"
 )
 
-func (s *service) ContactAliasKeySend(ctx context.Context, req *protocoltypes.ContactAliasKeySend_Request) (*protocoltypes.ContactAliasKeySend_Reply, error) {
+func (s *service) ContactAliasKeySend(ctx context.Context, req *protocoltypes.ContactAliasKeySend_Request) (_ *protocoltypes.ContactAliasKeySend_Reply, err error) {
+	ctx, _, endSection := tyber.Section(ctx, s.logger, "Sending contact alias key")
```
gfanton, line 14: Maybe we can set this kind of section inside a gRPC middleware ?

n0izn0iz: We could but we don't want it on all the calls so maybe not the best way

js/packages/store/context.ts
```diff
@@ -351,7 +362,10 @@ export const defaultPersistentOptions = (): PersistentOptions => {
 			value: 'performance',
 		},
 		[PersistentOptionsKeys.LogFilters]: {
-			format: 'info+:bty*,-*.grpc warn+:*.grpc error+:*',
+			format: '*:bty*',
+		},
+		[PersistentOptionsKeys.TyberHost]: {
+			address: Platform.OS === 'android' ? '10.0.2.2:4242' : '127.0.0.1:4242',
```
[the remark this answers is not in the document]
n0izn0iz: This is the default, there is a setting in devtools

go/pkg/bertyaccount/service_account.go
n0izn0iz: see this playground https://play.golang.org/p/ejXulLs39tt

go/framework/bertybridge/bridge.go
n0izn0iz: > we should definitely avoid the use of Any and only log useful field of the manager (if you really want to log this)
it's they way tyber was set to be initialized, I agree we should change it but I deemed it out of scope of this PR

tool/tyber/go/parser/header.go
n0izn0iz: I did not really touch this part, this is how tyber is initialized but most of the fields in this struct are not used, I will refactor it in another PR

go/framework/bertybridge/bridge.go
```diff
@@ -39,11 +44,18 @@ type Bridge struct {
...
 
 	lifecycleManager    *lifecycle.Manager
 	notificationManager notification.Manager
 }
 
+func randomID() string {
+	bytes := make([]byte, 256/8)
+	_, _ = rand.Read(bytes)
+	return base64.RawURLEncoding.EncodeToString(bytes)
+}
```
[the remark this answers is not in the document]
n0izn0iz: because it's self-explanatory code, but will merge it, the init will get refactored in another PR if we all agree

tool/tyber/go/go.mod
n0izn0iz: I tried updating astilectron and changing back to `/v2`, now I have a different error:
```
2021/05/21 08:43:05 Executing go build -ldflags -X "main.AppName=Tyber" -X "main.BuiltAt=2021-05-21 08:43:05.458300063 +0200 CEST m=+11.682350494" -X "main.VersionAstilectron=0.48.0" -X "main.VersionElectron=11.4.3" -o /home/norman/go/src/berty.tech/berty/tool/tyber/output/linux-amd64/binary berty.tech/berty/tool/tyber/go
2021/05/21 08:43:05 bundling failed: bundling for environment linux/amd64 failed: building failed: no required module provides package berty.tech/berty/tool/tyber/go; to add it:
	go get berty.tech/berty/tool/tyber/go
```
I couldn't find where I can override what package astilectron targets, not sure why it's not `/v2`

go/internal/logutil/tyber.go
n0izn0iz: Wouldn't this add 2 sec to the startup time of the app whenever tyber is not available ? (blocking)

tool/tyber/go/go.mod
aeddi: The problem with astilectron is also that it will use real path on the file system to build the electron app.
So if your gomodule path is `berty.tech/berty/v2/tool/tyber` and the path on your filesystem is `~/go/src/berty.tech/berty/tool/tyber` (without v2), it won't work.

tool/tyber/go/parser/header.go
aeddi: No need @n0izn0iz, I think the best thing to do is to wait until the `initutil` package is moved out of internal so we can remove this duplicated structure.

go/internal/logutil/tyber.go
gfanton: It will have the same behavior as `Dial` method, except this: 
> If the context expires before the connection is complete, an error is returned. Once successfully connected, any expiration of the context will not affect the connection. `

go/framework/bertybridge/bridge.go
aeddi: From my side I think it's good enough as it is at the moment. I used [this log](https://github.com/berty/berty/blob/1cc215104c4ec34138fffb2229f79573a6225a4f/go/internal/initutil/logger.go#L85) for the Tyber init (which already existed) when I made the v1.
It would just be necessary as @moul says that the mobile also goes through this for its init to unify, rather than dupplying this part with a "hack".

go/internal/logutil/tyber.go
n0izn0iz: yes so it waits for the connection to complete or 2 seconds, you remove the goroutine that does that in parallel so it indeed adds 2 sec of boot time I think

aeddi: This will be refactored in [this PR](https://github.com/berty/berty/pull/3298) (see point 6) anyway, so it will no be a problem anymore to dial Tyber few minutes after the start of the session.
We will be able to handle more usecase:
- Tyber host address is updated in app settings
- mDNS detected a new Tyber host on the network, display a popup on the app to ask user if he want to forward logs

tool/tyber/go/go.mod
moul: let's remove the `/v2` for now, I'll check later if I can unify the gomodule config

aeddi (approved):
LGTM 👍

PR #3271, opened 2021-04-22
feat: added reply and reactions to medias

sakul-budhathoki:
- Highlight scrollTo message
- Position fix for reaction view
- Utilize longPress to show reaction view

Note: Removed ```selectable={true}``` from UserMessageComponent as it block long press  to open Emoji popover.


zxxma:
Looks great, LGTM 👍 
(front Android & iOS review)

clegirar (approved):
Code LGTM 👍

PR #3291, opened 2021-04-29
feat: rework some flags

gfanton:
- `p2p.bootstrap`: ipfs bootstrap node, `:default:` will set ipfs default bootstrap node
- `p2p.dht`: dht mode, can be: `client`, `server`, `auto`, `autoserver`,
- `p2p.dht-randomwalk`: if `true` dht will have dht randomwalk enable


aeddi (approved):
LGTM 👍

PR #3304, opened 2021-05-04
fix: navigation crash issue

sakul-budhathoki:
The problem existed from my previous PR pressable avatar. 

Error: useNavigation couldn't find 
Avatars were working fine in Chat view as Chat Views are inside NavigationStack. Avatars have also used in notification components, where it can't find useNavigation; but have to use navigationRef to navigate. I have fixed it.


clegirar (approved):
LGTM 👍

zxxma:
LGTM 👍

PR #3354, opened 2021-05-28
fix: ios build for apple m1

gfanton:
- fix: remove lib-tor
- fix: deps
- fix: remove OpenSSL-Universal


js/ios/Podfile
```diff
@@ -9,7 +9,7 @@ target 'Berty' do
   use_react_native!(:path => config["reactNativePath"])
 
   # override OpenSSL-Universal with own openssl static lib
-  pod 'OpenSSL-Universal', :path => "./OpenSSL-Universal-Override.podspec"
+  # pod 'OpenSSL-Universal', :path => "./OpenSSL-Universal-Override.podspec"
```
aeddi, line 12: Please remove both this comment and the podspec

aeddi: I made a clean PR to add libtor back then, so we will only have to check it if we want to reintegrate it. ;)

js/Makefile
```diff
@@ -356,14 +356,10 @@ ios/Frameworks/Bertybridge.framework: $(bridge_src)
 	cd .. && $(GO) mod download
 	cd .. && $(GO) install golang.org/x/mobile/cmd/gobind
 	cd .. && $(GO) run golang.org/x/mobile/cmd/gomobile init
-	# Download required go-libtor deps if needed
-	../tool/build/shell/dl-go-libtor-ios.sh
 	mkdir -p ios/Frameworks
```
aeddi, line 359: In this file you should also remove:
- `rm -rf ios/tor-deps` line 115
- `ios/OpenSSL-Universal-Override.podspec` deps of `ios/Berty.xcworkspace` rule on line 373

aeddi: Can you also remove the file `tool/build/shell/dl-go-libtor-ios.sh`?

aeddi (changes requested):

aeddi (approved):
LGTM 👍

PR #3356, opened 2021-05-30
fix: betabot send date and quick reply options

sakul-budhathoki:
**Issues**
1. Reply Options is not shown in betabot chat view  due to error: ```[Error: invalid wire type 4 at offset 5]```
2. Send date is not shown in betabots due to ```nextMessage``` is the ```inte``` with ```payload.options``` which is quick reply options.

**Solution**
I have filtered out the message which have ```payload.options``` to fix send data. And for reply option; extract the last message which has ```payload.options``` and use it for quick reply options

**Questions**
1. Was adding quick reply options ```payload.options``` in messages is intentional or bug?  
2. Instead of extracting payload.options Can we fix the above error?

### DIFF
--- a/js/packages/components/chat/MessageList.tsx
+++ b/js/packages/components/chat/MessageList.tsx
@@ -63,7 +63,11 @@ export const MessageList: React.FC<{
 	const conversation = useConversation(id)
 	const ctx = useMsgrContext()
 	const members = ctx.members[id]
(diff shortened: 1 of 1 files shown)


zxxma:
(Front review, looks good to me 👍)

n0izn0iz (approved):

PR #3348, opened 2021-05-25
fix(messenger): remove useless locks

n0izn0iz:
bench using Interact on multiple node in parallel with a TestingInfra (aka sending messages in spam conditions)
```
nodes: 10
messagesPerNode: 300
messagesTotal: 3000
```
1513081f0a43a11f26998ec69bea678e9cc359d9 (master)
```
avg: 8m4.866547189s
min: 7m24.719786834s
max: 8m49.098070929s
avg/msg: 1.616221823s
min/msg: 1.482399289s
max/msg: 1.763660236s
```
9747980998f6ab69d6ce41eda7866750397b0297 (this branch)
```
avg: 2m29.030747214s
min: 2m20.090942079s
max: 2m35.32906842s
avg/msg: 496.769157ms
min/msg: 466.969806ms
max/msg: 517.763561ms
```
summary: more than 3 times faster without the lock

PR #3362, opened 2021-06-02
WIP: fix: remove non atomic locks, optimize datastore interactions

aeddi:
In short, the slow sending of messages on Berty seemed to come from the 3 main layers:
- Messenger: locks with a much too large scope that prevented parallelizing api calls (fixed by #3348)
- Protocol: too large locks in message_keystore and a badly optimized put in the datastore (fixed by this PR)
- Orbit-DB: the time of an `AddOperation` is abnormally long, we need to find the cause and fix.


aeddi:
Fixed by #3319

PR #3376, opened 2021-06-10
fix: background fetch

gfanton:
potentially fix background fetch crash where service account was not initialized yet.
add a protection over background fetch exec so it will return an error instead
of crashing if service account is not initialized.


aeddi (approved):
LGTM 👍

PR #3379, opened 2021-06-13
fix: android recording error

sakul-budhathoki:
App crash due to error from `react-native-audio-toolkit` package
PR created: https://github.com/react-native-audio-toolkit/react-native-audio-toolkit/pull/264


n0izn0iz (approved):
code lgtm

PR #3397, opened 2021-06-23
fix: prevent proguard to strip inAppBrowser required symbols

aeddi:


clegirar (approved):
LGTM 👍

PR #3407, opened 2021-06-28
feat: tyber discovery + removed tracer package

glouvigny:


js/packages/store/providerEffects.ts
```diff
@@ -179,6 +178,14 @@ export const openingDaemon = async (
 		return
 	}
 
+	let tyberHost = ''
+	try {
+		tyberHost = (await AsyncStorage.getItem(tyberHostStorageKey)) || ''
+		console.warn(`connecting to ${tyberHost}`)
```
aeddi, line 184: Shouldn't you check if `tyberHost != ''` before logging that?
I presume it could be the case because of this line:
`tyberHost = (await AsyncStorage.getItem(tyberHostStorageKey)) || ''`

aeddi (approved):
LGTM 👍

PR #3412, opened 2021-06-30
fix: unset Android env var to prevent gomobile error on init

aeddi:


gfanton (approved):
LGTM 👍

PR #3413, opened 2021-06-30
fix: unset Android env var to prevent gomobile error on bind

aeddi:
Forgot to add this change in #3412 


clegirar (approved):
LGTM 👍

PR #3425, opened 2021-07-09
feat: add BertyAccount.ImportWithProgress alternative endpoint

moul:


moul:
@clegirar can you confirm me that it works as expected?

clegirar (commented):
LGTM 👍 (test only)

PR #3429, opened 2021-07-11
 permission view

sakul-budhathoki:


js/packages/components/main/Permissions.tsx
```diff
@@ -0,0 +1,251 @@
...
+		}
+		if (navigateNext) {
+			props.navigation.navigate(navigateNext, {})
+		} else {
+			props.navigation.goBack()
+		}
+	}
+
+	const handleAppStateChange = async (nextAppState: string) => {
+		if (appState.current.match(/inactive|background/) && nextAppState === 'active') {
```
n0izn0iz, line 53: could you add a comment with the rationale for checking permissions on AppState change pls? can't figure it out

n0izn0iz (approved):
LGTM

js/packages/components/main/Permissions.tsx
sakul-budhathoki: Ok. If the permission is denied, then we will show user "Go to Setting" button to let them enable permission from Settings. After enabling it from settings, user will be back to our app.At that moment, we will check if permission is enabled or not using AppState change; if enabled we will navigate back. If not user will be on same screen. I hope it's clear now.

zxxma:
Front (iOS) : ✅  LGTM

PR #3470, opened 2021-07-30
fix: BLE crashes

D4ryl00:
# iOS
* change BleManager singleton for non-ARC version (crashed when switching account)
* stop scanner timer in the main thread (crashed when stopping the node)

# Android
* Track correctly the broadcast receiver register status to avoid to unregister it multiple times (crashed when switching account)


aeddi (approved):
LGTM 👍

PR #3475, opened 2021-08-03
fix: unilateral connections on BLE driver

D4ryl00:
Each device has two roles: advertiser and scanner (central and peripheral on iOS).
Now, to initialize a Berty connection, a device is now only a client or server, not both at the same time as previously.


go/internal/ble-driver/BertyDevice_darwin.m
```diff
@@ -333,65 +379,90 @@ - (void)peripheral:(CBPeripheral *)peripheral didOpenL2CAPChannel:(CBL2CAPChanne
...
-                }
-            } else {
-                os_log_debug(OS_LOG_BLE, "🟡 didUpdateValueForCharacteristic: peerID unknown in connectedPeers");
-                ConnectedPeer *peer = [[ConnectedPeer alloc] init];
-                [peer setClient:self];
-                [peer setClientReady:TRUE];
-                [PeerManager addPeer:peer forPeerID:remotePeerID];
-            }
-        });
+        [self.writeQ completedTask:error];
```
aeddi, line 460: If this line means "completedTask without error" can you change it for `[self.writeQ completedTask:nil];`.
It's very unclear that error is nil at this point.

Also, is it normal that the various `os_log_debug` that log an error (lines 457 and 464) do not call this callback with an error?

aeddi (approved):
LGTM 👍

PR #3486, opened 2021-08-09
fix: bump libs

clegirar:


zxxma:
LGTM 👍  (iOS & Android UX review)

js/patches/react-native+0.63.4.patch
```diff
@@ -1,30 +0,0 @@
...
-+## @NOTE(gfanton): patch config.sub to work with apple silicon
-+# download patched config.sub for darwin arm64
-+curl -sSL --fail -o config.sub 'https://gist.githubusercontent.com/gfanton/f7dd19c299bef99d323cb08c650bb02e/raw/2f5a76ffd38cdac678adf6a9f12600e9d16e7345/config.darwin-arm-patch.sub'
-+
-+# check shasum of the file
-+echo 'a743811464b00f3e2372f5368035fdc38f322a3a  config.sub' | shasum -a 1 -c || exit 1
-+
- ./configure --host arm-apple-darwin
- 
- # Fix build for tvOS
```
aeddi, line 30: Are you sure we don't need this patch from @gfanton?
Did you try to compile the iOS app from XCode on both debug and release from a x86 and a M1 Mac?

aeddi (approved):
LGTM 👍

js/android/build.gradle
```diff
@@ -2,18 +2,18 @@
...
 buildscript {
     ext {
-        buildToolsVersion = "29.0.2"
-        minSdkVersion = 23
+        buildToolsVersion = "29.0.3"
+        minSdkVersion = 21
         compileSdkVersion = 29
         targetSdkVersion = 29
-        ndkVersion = "22.1.7171670"
+        ndkVersion = "20.1.5948944"
```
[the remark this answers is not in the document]
clegirar: It was recommended by react-native-upgrade-helper but yes i can keep current master version

js/patches/react-native+0.63.4.patch
clegirar: Yes, this patch was for recognize M1 like a valid arch (for flipper that use an old version of glog) but now it's detect automatically, so the builds (x86 and arm64) works 👍

zxxma:
LGTM 👍

PR #3498, opened 2021-08-19
feat: trash animation

sakul-budhathoki:


zxxma:
Looks great @sakul-budhathoki !

We need to fix two little bugs before merge : 

1) The little helper text should be a little bit higher, to not be hidden behind txt input
image

2) There is a little glitch on the animation when we delete the audiofile, the camera & microphone buttons are "struggling" before going back to their original place. I guess it's related to the slide movement when we put the record into the trash, but I don't know exactly how the glitch could be cleaned.

zxxma:
This PR will need to a big rebase, so we post-pone it for the moment!

PR #3501, opened 2021-08-20
fix: kanban fixes

clegirar:
#### Some fixes
- [x] Fix police color of account names in CreateGroup
- [x] Continue text button in CreateGroup was bad aligned
- [x] Add Appearence in Settings (remove from DevTools)
- [x] Verify that OS language is automatically detected and set on the App
- [x] Ask for permission AFTER camera allowing request
- [x] Add backgroundColor of the QR Code component
- [x] Clean i18n en-US json file
---
#### Delete WelcomeModal
- [x] Desactivate welcome modal and add a default conversation to ask to enable replication service
---
#### Chat issues
- [x] Save textInput when we go back
- [x] Fix KeyboardAvoidingView in conversations
- [x] Remove Betabot reply options
---
#### CreateAccount issues
- [x] Add a loader instead of text input when the account is creating
- [x] Fix the kind of jump when we navigate to SetupFinished
---
#### Permission issues
- [x] Refracto permission


zxxma:
Bug found: After arriving in Home, Replication Services Config conversation is clickable, but is locked in a loop and doesn't allow to confirm it's done. Since this moment, even when we click on Notification Configuration, it open&close fastly the Replication service.

clegirar:
Yep this bug is here for a long time, i can dig it but it doesn't happen every time, hard to reproduce

clegirar:
We have a problem with the reset of the navigation (https://github.com/react-navigation/react-navigation/issues/9872), i got a fix in waiting for an answer

PR #3575, opened 2021-10-04
fix: conversation related issues

glouvigny:
db: GetAllConversations() was returning multiple entries for a single conversation
app: lazy loading of messages on conversation opening and logic welcome banner
messenger: subscribe to groups after creating/joining groups


gfanton (approved):
LGTM 👍

PR #3651, opened 2021-11-10
fix: microphone button on Android

### LINKED ISSUE #3648: [BUG]: Microphone button never triggered on Android
### Actual Results / Problem

When I press the microphone button in the chat, nothing happens

### Expected Results / Possible Fix

I expect the microphone button to be triggered

### Give use some context

This bug only happens on Android

JohnBerd:
react-native-gesture-handler does not trigger callbacks without those configuration changes

fix #3648


n0izn0iz (approved):
LGTM

PR #3686, opened 2021-11-15
fix: notification badge below the avatar

JohnBerd:
Notification badge was behind the avatar


n0izn0iz:
does not seem to fix the problem on samsung a10s

JohnBerd:
> does not seem to fix the problem on samsung a10s

You are right, I had a bug with fast refresh

n0izn0iz (approved):
LGTM

PR #3698, opened 2021-11-17
fix: disalbe `embed_swift` for Notification Service

gfanton:
we need to set this to false to avoid a `Validation Error` while submitting
the build to apple


aeddi (approved):
LGTM 👍

PR #3700, opened 2021-11-18
fix(ci): re-enable checking Podfile.lock sync

n0izn0iz:
this will prevent merging js bump PRs without updating the deps native part


aeddi (approved):
LGTM 👍

Issue #3393, opened 2021-06-21
[BUG]: QR code scan even when popup is there


KristofKekesi:
### Actual Results / Problem

When I'm scanning a QR code to join a group with an invalid QR (edit: also with a valid QR)  the popup will show up, but the program is still looking for QRs. It wouldn't be noticable if my phone wasn't shaking.

### Expected Results / Possible Fix

Don't scan for QR if popup is there.

### Give use some context

![198164834_860393541221668_4581254370981712286_n](https://user-images.githubusercontent.com/64812024/122841076-b29a1a00-d2fb-11eb-94fc-1828df8a9e08.jpg)

### If you are running on mobile, what's your version/branch? what's your device and OS?
I'm on iOS 14.7 (29G5023c) with an iPhone 11, The version of Berty is 2.42.1.3.

### Anything else you want to tell us, like your favorite cheese?
Well if you ask :D it is  mozarella.

aeddi:
Hello, thanks for your input! 👍 
We'll look into it and we'll update this issue when it's fixed.

Issue #3447, opened 2021-07-20
[BUG]: Repeat QR code reads


OliverBrotchie:
### Actual Results / Problem

Berty will rapidly and repeatedly scan and vibrate whilst held over a QR code.

### Expected Results / Possible Fix

Whilst Berty should constantly look for QR codes through the camera if it has already found a code it shouldn't continuously vibrate if the QR code is the same as the code displayed on-screen.

### Give use some context

This happened whilst scanning the QR codes provided in the #beta-qr-sharing discord channel.

Phone: One Plus Nord - AC2003
Os: Oxygen OS 11.1.1.3.AC01BA
Berty: v2.293.6

LFGaming:
This bug is known yes, it’s a duplicate of [Bug #3393](https://github.com/berty/berty/issues/3393)

OliverBrotchie:
Ah okay, thanks!

PR #3713, opened 2021-11-23
fix: pause QR code scanner

JohnBerd:
QR code gets hidden after a QR code scanning
QR codes can successfully be scanned several times


n0izn0iz:
I'm not sure creating a reducer for this is the good tool, can't this be done with the navigation blur/focus hooks?

JohnBerd:
> I'm not sure creating a reducer for this is the good tool, can't this be done with the navigation blur/focus hooks?

Without reducer, the solution remaining would be to pass a callback to route params when navigating to ManageDeepLink component. If it seems to be a better solution for you I can go for it :)

n0izn0iz (approved):
code LGTM

clegirar:
Test LGTM 👍

JohnBerd:
fix #3393, #3447

PR #3715, opened 2021-11-24
feat: add golang i18next support

gfanton:


aeddi (commented):
Did you forget to modify `go/framework/bertybridge/push.go`? Because as it is, I don't see how background push notifs can be formatted on Android.

aeddi (approved):
LGTM 👍

PR #3741, opened 2021-12-08
fix: apply OS language

clegirar:


zxxma:
Looks good for the feature, but the Language is not displayed on the Setting List.
Looks like this : 

image

PR #3743, opened 2021-12-09
fix: onboarding create account name screen

glouvigny:
* Link device was not working and is not currently implemented, and has been removed
* More info wasn't working either, and has been removed
* Advanced in the dropdown should be Import
* We had issue with alignment in the info box at the bottom, I removed the info icon

image

image


aeddi (commented):
LGTM 👍

aeddi (approved):

PR #3747, opened 2021-12-10
fix: add LF dark theme in default collection theme

clegirar:


zxxma:
LGTM 👍 
Thanks LF for contribution 🥇

zxxma:
@clegirar may we add it as main default darktheme, and keep the purpledark as choice in the theme lib? :)

PR #3754, opened 2021-12-12
fix: rename audio messages and allow sharing them

n0izn0iz:


zxxma:
Sounds Good, audio rec are in Documents and well renamed 
gg bro

Just a fix on soundwave is needed

zxxma:
LGTM 👌

PR #3760, opened 2021-12-13
fix: cancel reply cross is outside box with long large text

clegirar:
Fix https://github.com/orgs/berty/projects/42#card-74695232

signal-2021-12-13-185016
signal-2021-12-13-185021


zxxma:
LGTM 👍

PR #3769, opened 2021-12-14
fix: can collapse long message after click on read more

clegirar:


zxxma:
iOS iPhone & iPad review 
LGTM 👍

PR #3770, opened 2021-12-14
fix: display qr code in right size on ipads

JohnBerd:
Signed-off-by: Xavier <lecunffxavier@gmail.com>


zxxma:
Bug detected: I have a glitch when the QR Code is displayed. After the glitch, sizing is ok, but during 1sec, it's glitched.

n0izn0iz (approved):
LGTM, maybe move the hook to the styles package but can do in another PR

n0izn0iz (changes requested):
the glitch need to be fixed

zxxma:
Glitch fixed ✅

PR #3778, opened 2021-12-15
fix: pass media metadata to network

n0izn0iz:


zxxma:
We detected a little bug, really strange, one of audio rec was kind of repeated 2 times.
UI was displaying 0'9 sec, but file was playing 18 sec. 
But we cannot reproduce, and all others tests were perfect.
So, n1 bro.

LGTM 👍

PR #3783, opened 2021-12-16
fix: cheatcode enter key on ipad external physical keyboard

clegirar:


n0izn0iz (changes requested):
we need a proper fix for this dos

PR #3789, opened 2021-12-16
fix: description error screens

clegirar:
Fix https://github.com/orgs/berty/projects/42#card-74892976


zxxma:
LGTM 👍

PR #3792, opened 2021-12-16
fix: protect against nil payloads

n0izn0iz:


zxxma:
Will need another dig session about some specific inputs I tried, but useful PR too, reviewed.
LGTM 👍

PR #3796, opened 2021-12-16
WIP - feat: improve file menu and use generic cancel internationalization

n0izn0iz:


n0izn0iz:
superseded by #3803 until we have real multimedia messages
we should extract the generic cancel i18n part

PR #3815, opened 2021-12-21
Home redirection

JohnBerd:


zxxma:
need to point to the conversation (when scanning an already scanned QR)

zxxma:
LGTM 👍

PR #3832, opened 2021-12-29
feat: improve audio message playback ux

n0izn0iz:


zxxma:
LGTM 👍 
iOS review (iPad+iPhone) + Darktheme versions

PR #3835, opened 2021-12-30
feat: make messenger and protocol inits standalone

moul:

### DIFF
--- a/api/accounttypes.proto
+++ b/api/accounttypes.proto
@@ -16,16 +16,13 @@ service AccountService {
   /// Account
 
   // OpenAccount starts a Berty node.
-  rpc OpenAccount (OpenAccount.Request) returns (OpenAccount.Reply);
-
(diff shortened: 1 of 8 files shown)


go/pkg/bertyaccount/service.go
```diff
@@ -112,10 +109,13 @@ func (o *Options) applyDefault() {
...
+func NewService(opts *Options) (Service, error) {
 	rootCtx, rootCancelCtx := context.WithCancel(context.Background())
 
-	var s *service
+	var (
+		s   *service
+		err error
+	)
 	endSection := tyber.SimpleSection(rootCtx, opts.Logger, "Initializing AccountService")
 	defer func() { endSection(err, tyber.WithDetail("RootDir", s.rootdir)) }()
```
n0izn0iz, line 120: removing the named return will prevent tyber from catching the correct error

PR #3837, opened 2021-12-31
fix: don't dispatch useless member updates

n0izn0iz:
also wraps multiMemberGroupInitialMemberAnnounced handler in tx


jefft0:
n0izn0iz says: "This PR could still be relevant. This is an optimization so it doesn't trigger "state changes" and thus logic + rerenders in the front. But it's not very important I believe."

jefft0:
Keep open because of the comment from n0izn0iz. Not a high priority, and not a bug. Put in backlog.

PR #3843, opened 2022-01-03
fix: re-add react-native patch for apple silicon

gfanton:
fix apple m1 native `pod install`


aeddi:
Tested today on master without these changes:
- `cd js && make clean`
- `rm -rf $HOME/.gradle/caches                                                                                                                                  Native $HOME/Library/Developer/Xcode/DerivedData $HOME/Library/Caches/CocoaPods`

I can still build and run the application on both device and simulator from a M1 Mac (native without Rosetta).
This change seems now obsolete, so I close this PR.

PR #3852, opened 2022-01-08
fix: crash when iOS BLE driver stops

D4ryl00:

Fix a crash when the iOS BLE driver is stopping.
Fix a stuck when the iOS BLE driver starts on simulators.


aeddi (commented):
LGTM 👍

aeddi (approved):

PR #3875, opened 2022-01-21
feat: add emoji suggestion modal

JohnBerd:
- a suggestion modal appears when typing an emoji shortcode (eg: :wa ve: without space :))
- shortcode in text inputs are automatically replaced by the corresponding emoji
- theme is now typed
- blue offset in home on IOS is removed
- chat buttons are aligned with the input

### DIFF
--- a/js/packages/berty-i18n/locale/en-US/messages.json
+++ b/js/packages/berty-i18n/locale/en-US/messages.json
@@ -469,35 +469,35 @@
 			"fingerprint": "Fingerprint",
 			"info": "Infos",
 			"devices": "Devices",
-      "my-devices": "My devices"
+			"my-devices": "My devices"
 		},
 		"multi-member": {
 			"input-placeholder": "Type here..."
 		},
-    "push-notifications": {
-      "enable": "Enable push notifications",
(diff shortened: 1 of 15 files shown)


zxxma:
Bug detected:
- Looks like the modal comes over the conversation
- At conv opening we can see the modal swiping up quickly over the conv, and when we send a msg, everything is hidden by the modal

js/packages/redux/reducers/chatInputs.reducer.ts
```diff
@@ -12,15 +12,23 @@ const makeRoot = <T>(val: T) => ({
...
+	isSending: boolean
 	mediaList: string[] // cids
 }
 
 const newChatInputState: (convPK: string) => ChatInputState = convPK => ({
 	id: convPK,
 	text: '',
+	selection: null,
+	isFocused: false,
+	isSending: false,
```
n0izn0iz, line 31: are you sure you want to persist state that will not be in sync after an app restartt

n0izn0iz (approved):
LGTM

zxxma:
LGTM 👍

PR #3888, opened 2022-02-03
fix: fix scroll problems on Android messages in conversations

JohnBerd:
- Scroll more natural on Android
- Spinner is displayed if messages are loading from Go
- Scrolling with finger on an other user’s message should work even if scrolled in diagonal
- Reply by scrolling right on an other user’s message should still work


zxxma:
LGTM 👍  (Android review)

n0izn0iz (approved):
LGTM

PR #3889, opened 2022-02-03
feat: jump to a group after accepting group invitation

JohnBerd:


clegirar (commented):
You have two console.log remaining but idk if we want to keep them, else the code seems good 👍

clegirar (changes requested):
Test LGTM 👍

PR #3892, opened 2022-02-05
fix: request camera and gallery permissions in chat menu footer

JohnBerd:
This PR makes the user navigate to a permission screen after pressing on gallery and camera chat footer buttons (after clicking on "+") if he doesn't have those permissions.

It could need a lottie file for gallery permission, camera lottie file is temporary used instead.

The behavior can still be: asking for permission without navigating to permissions screen.

### before

permissionBefore

### after

permissionafter

### DIFF
--- a/js/packages/berty-i18n/locale/en-US/messages.json
+++ b/js/packages/berty-i18n/locale/en-US/messages.json
@@ -622,6 +622,10 @@
 			"title": "Audio messages",
 			"desc": "To send audio message to your contacts, you need to allow microphone permission to Berty app."
 		},
+		"gallery": {
+			"title": "Image messaging",
+			"desc": "To send image message to your contacts, you need to allow gallery permission to Berty app."
+		},
 		"button-labels": {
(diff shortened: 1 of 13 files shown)


clegirar (commented):
Code LGTM 👍

clegirar (changes requested):
I just tested the app and if I accept the permission, i don't navigate to the camera or the gallery directly
I think the behavior that we want it's:
- If the user accept the permission, we'll navigate directly to the camera or the gallery
- If the user deny the permission, we just go back like the behavior actually

So for that check if the `onPress` function doesn't return before run the camera or the gallery

---

And an other thni maybe you could add a `setActiveTab(TabItems.Default)` before the return of each `onPress` of each element of the `LIST_CONFIG`

JohnBerd:
The changes requested hasn't been done on this request, for code refactoring purpose, it will be simpler to fix them in https://github.com/berty/berty/pull/3910

PR #3895, opened 2022-02-08
feat: interaction's reactions

JohnBerd:
- reaction counter on a message
- button to add a reaction if a reaction is already on a message
- reactions of the current user are colored
- a long press on a reaction displays the list of user having the same reaction on a message

### before
testgifbefore

### after
testgif


clegirar (changes requested):
Except the memoized functions and the log, the code seems good (i don't test the PR, @zxxma did it), so LGTM after changes 👍
Maybe just change the PR's name :)

js/packages/components/shared-components/AnimatedNumber.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+			Animated.timing(animation, {
+				toValue: -1 * (numberHeight * animateToNumbersArr[index]),
+				duration: animationDuration || 1400,
+				easing: easing || EasingNode.elastic(1.2),
+			}).start()
+		})
+	}, [number, animateToNumbersArr, animationDuration, animations, easing, numberHeight])
+
+	const getTranslateY = (index: number) => {
```
clegirar, line 57: Sorry i forgot this function, same for the memoized hook, i think we have to use a `useCallback` here to

JohnBerd: Thank you for the constructive review <3

clegirar (approved):
LGTM 👍

PR #3899, opened 2022-02-10
fix: fix crash on MacOS with the BLE driver

D4ryl00:
The crash is caused by CBCentralManager and CBPeripheralManager which don't allow a custom dispatch queue.


D4ryl00:
It's a system bug where we need to add or remove iTerm is system preference -> security & privacy -> Privacy -> Bluetooth

PR #3907, opened 2022-02-15
fix: improve and fix integration test

gfanton:
- fix integration test
- speed up integration test
- add a log when a peer is found but was filtered


aeddi (approved):
LGTM 👍

PR #3908, opened 2022-02-16
chore(ci): ruby pinning and simpler caching

n0izn0iz:
this prepares the react-native upgrade which requires a pinned ruby and uses a `.ruby-version` at the app root (currently still in `js/ios`)


aeddi (approved):
LGTM 👍

PR #3910, opened 2022-02-17
feat: conversation modal logic

JohnBerd:
Rebased over #3895

- New conversation modal logic with a context
- Remove Popover and reply context (logic moved to redux)
- Message interaction popover => message interaction modal

### before
popupbefore

### after
popupafter


JohnBerd:
Moved to #3973

PR #3912, opened 2022-02-18
fix(ci): replace setup-ruby on self-hosted

n0izn0iz:


aeddi (changes requested):
We fixed ruby on the self hosted in this PR #3919 (I also had to do some modifications on the runners). It works fine now, but we didn't cache the gems.

PR #3916, opened 2022-02-18
chore(ci): disable self-hosted runners until fixed

n0izn0iz:


moul (changes requested):
I just updated the runners, can you try again?

moul:
Merging to fix the CI, let's see later how to safely reactive the dc4 runners

PR #3935, opened 2022-02-28
feat: increase size of app font

JohnBerd:
- increase font size
- remove most of hardcoded fontSize properties
- type useStyles for autocompletion
- type useThemeColors for autocompletion


clegirar (approved):
Code LGTM 👍

PR #3956, opened 2022-03-07
feat: add share modal replacing QR scan and invitation home floating …

JohnBerd:
### before
sharebefore

### after
shareafter


clegirar (changes requested):
On Android we have many issues:
- The "plus" button in Home, is not center and too big


- The status bar in the Share screen is not visible
- The Create group button does not work

aeddi:
Duplicate #3960

PR #3958, opened 2022-03-08
fix: ios background crash

D4ryl00:
Use patched repo of berty.tech/go-ipfs-repo-encrypted which fixes a crash when iOS app goes in background mode.

depends on https://github.com/berty/go-ipfs-repo-encrypted/pull/22

### DIFF
--- a/go/internal/accountutils/utils.go
+++ b/go/internal/accountutils/utils.go
@@ -38,6 +38,8 @@ const (
 	ReplicationDatabaseFilename = "replication.sqlite"
 	StorageKeyName              = "storage"
 	StorageKeySize              = 32
+	StorageSaltName             = "storage_salt"
+	StorageSaltSize             = 16
 )
 
 func GetDevicePushKeyForPath(filePath string, createIfMissing bool) (pk *[cryptoutil.KeySize]byte, sk *[cryptoutil.KeySize]byte, err error) {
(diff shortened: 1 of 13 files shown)


go/internal/accountutils/utils.go
```diff
@@ -129,7 +139,7 @@ func ListAccounts(ctx context.Context, rootDir string, ks NativeKeystore, logger
 
 var storageKeyMutex = sync.Mutex{}
 
-func getOrCreateStorageKey(ks NativeKeystore, keyName string) ([]byte, error) {
+func getOrCreateKeystoreKey(ks NativeKeystore, keyName string) ([]byte, error) {
```
gfanton, line 142: can you add keysize as argument?

gfanton (changes requested):
can you just update the size of the salt?
also it could be great to create a `struct` that embed the salt and the key to avoid adding both arguments to every function that need it (it should be ok for now has we need this fix as fast as possible)

go/internal/accountutils/utils.go
```diff
@@ -37,6 +37,7 @@ const (
 	MessengerDatabaseFilename   = "messenger.sqlite"
 	ReplicationDatabaseFilename = "replication.sqlite"
 	StorageKeyName              = "storage"
+	StorageSaltName             = "storage_salt"
 	StorageKeySize              = 32
```
gfanton, line 41: Can you add a `StorageSaltSize` const?

go/internal/accountutils/utils.go
```diff
@@ -153,19 +163,27 @@ func getOrCreateStorageKey(ks NativeKeystore, keyName string) ([]byte, error) {
...
-	return getOrCreateStorageKey(ks, fmt.Sprintf("%s/%s", StorageKeyName, accountID))
+	return getOrCreateKeystoreKey(ks, fmt.Sprintf("%s/%s", StorageKeyName, accountID))
+}
+
+func GetOrCreateMasterStorageSalt(ks NativeKeystore) ([]byte, error) {
+	return getOrCreateKeystoreKey(ks, StorageSaltName)
+}
+
+func GetOrCreateStorageSaltForAccount(ks NativeKeystore, accountID string) ([]byte, error) {
+	return getOrCreateKeystoreKey(ks, fmt.Sprintf("%s/%s/salt", StorageKeyName, accountID))
```
gfanton, line 178: can you update the salt size based on signal's YapDatabase, they have to do same fix has us and it's look like they use a 16bytes salt https://github.com/signalapp/YapDatabase/pull/1/files#diff-550bf64a8af9dd55eca1060a2d9fe142b068e5c01925326e9d22ccb2714febe6R94

gfanton (approved):
LGTM 👍  dont forget to update the `go.mod` after merging: https://github.com/berty/go-ipfs-repo-encrypted/pull/22

PR #3967, opened 2022-03-15
fix: push notif button

clegirar:
I have implemented a bad behavior of render on this component, and it broke on iOS


D4ryl00 (approved):
LGTM 👍
Tested on iOS and Android.

PR #3886, opened 2022-02-01
WIP: chore: add react native web

clegirar:
Signed-off-by: clegirar <clemntgirard@gmail.com>

PR #3979, opened 2022-03-18
ci: add a lib check dependencies job

JohnBerd:


aeddi (approved):
LGTM 👍

aeddi (changes requested):
After doing a comparison with the manual cleanup done in [this commit](https://github.com/berty/berty/pull/3886/commits/dfdaaeca6074409acbd6a7e69f198a5b958105de) a while ago, I think depcheck is really not reliable enough to rely on (many false positives and false negatives). 
On the other hand, I didn't remove react-native-flags and I think it's a good idea looking at your changes. Can you keep only this removal and rename your PR?

PR #3990, opened 2022-03-22
fix: order imports on JSX files

JohnBerd:


clegirar:
We have a lot of conflict, more simple to just create a new branch 👍

PR #3992, opened 2022-03-22
fix: enable localdiscover

D4ryl00:
Enable the localdiscover driver for tinder to allow contact request between two local peers.


gfanton (approved):
LGTM 👍

PR #4025, opened 2022-03-28
fix: some issues on web

clegirar:
- [x] goBack arrows on web are not visible
- [x] Unify font rendering by using Open Sans everywhere
- [x] Support Open Sans font on web
- [x] Support opening berty.tech/id(http://berty.tech/id) URLs

### DIFF
--- a/js/packages/components/shared-components/BText.tsx
+++ b/js/packages/components/shared-components/BText.tsx
@@ -1,0 +1,22 @@
+import { useThemeColor } from '@berty-tech/store'
+import { useStyles } from '@berty-tech/styles'
+import React from 'react'
+import { Text, TextProps } from 'react-native'
+
+export const BText: React.FC<TextProps> = props => {
+	const { children, style } = props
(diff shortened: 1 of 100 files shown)


js/packages/components/shared-components/BText.tsx
```diff
@@ -0,0 +1,22 @@
+import { useThemeColor } from '@berty-tech/store'
+import { useStyles } from '@berty-tech/styles'
+import React from 'react'
+import { Text, TextProps } from 'react-native'
+
+export const BText: React.FC<TextProps> = props => {
```
aeddi, line 6: I think you should rather choose between these two options to name this component:
1. `Text` so that you can't import both our component and react-native's one, the goal being I guess to use only ours starting now
2. if the first option is not desirable, choose a more descriptive name like `UnifiedText` or something like that, but not `BText` please 😉

glouvigny (approved):

PR #4038, opened 2022-03-31
fix: persist of persitent options

clegirar:


clegirar:
We have some issues with the go datastore that saved js options
We will start soon a big refacto and the js options is in, so it will be fix in this story 👍

PR #4041, opened 2022-04-01
feat: enable HolePunching and expose it as a flag

Jorropo:
This has been added in go-ipfs `v0.11.0` and will be defaulted to true in the soon to be released `v0.13.0`, you can safely enable this. (the reason it wasn't enabled by default is that the network first needed enough people to upgrade their nodes to a version that supports Circuit V2).

It has works more often than not efficiency in go-ipfs. It helps with non-symmetric nats that doesn't support UPNP (which covers most home networks). Many cellular carrier use symmetric nats but there is also a non negligible amount that use hole punchable NATs so it should help there too.

You can read more about it here: https://github.com/libp2p/specs/blob/master/relay/DCUtR.md

### Mobile:

I forgot how to do options to mobiles.

I don't think it's needed to expose that to the users.
But on mobile it should default to `true` (that is taken careoff by the flag framework right now which I recall if it runs on mobile or not).


Jorropo:
I have no idea what is this CI failure, looks like golangci/golangci-lint#2374, I use go-pls which runs gofumpt on everything.

gfanton:
@Jorropo thanks a lot ! We need to test this on cellular first to see if it's working and if we gain some performances.  
(Go / GolangCI-lint has an issue we go 1.18, i fixed this on master)

Jorropo:
We are enabling it by default in go-ipfs v0.13.0 this isnt needed.

PR #4054, opened 2022-04-07
feat: move assets in one folder

JohnBerd:


js/Makefile
```diff
@@ -79,6 +80,11 @@ generate: node_modules gen.sum
 .PHONY: regenerate
 regenerate: gen.clean generate
 
+.PHONY: icons.generate
+icons.generate:
+	npx icon-font-generator packages/assets/custom-icons/*.svg -o packages/assets/custom-icons/font --name custom-icons-font
```
clegirar, line 85: Here you can use something like that to check if npx is a callable program: `$(call check-program, npx)`

js/packages/components/chat/ChatSettingsMemberDetail.tsx
```diff
@@ -64,7 +65,7 @@ export const ChatSettingsMemberDetail: ScreenFC<'Group.ChatSettingsMemberDetail'
 	React.useLayoutEffect(() => {
 		navigation.setOptions({
 			title: displayName,
-		})
+		} as Partial<StackNavigationEventMap>)
```
clegirar, line 68: I don't know why we have to cast this object now, try to remove it and run `make lint` (that check the typescript errors) to see if we really have an error

js/packages/messenger-app/custom-icons.tsx
```diff
@@ -1,48 +1,48 @@
...
-import Settings from './custom-icons-svgs/settings.svg'
-import Config from './custom-icons-svgs/config.svg'
-import Proximity from './custom-icons-svgs/proximity.svg'
-import Peer from './custom-icons-svgs/peer.svg'
-import Services from './custom-icons-svgs/services.svg'
-import ExpertBluetooth from './custom-icons-svgs/expert-bluetooth.svg'
-import ExpertSetting from './custom-icons-svgs/expert-mdns.svg'
-import ExpertPushNotif from './custom-icons-svgs/expert-push-notif.svg'
-import ExpertNode from './custom-icons-svgs/expert-node.svg'
+import Fingerprint from '@berty/assets/fingerprint.svg'
```
clegirar, line 4: Can you check in this file if all custom-icons are used in the app, if someones are not used just delete it please 👍

clegirar (changes requested):
- The package-lock.json should not exsit, please delete it
- I think you have to delete all ttf files and convert svg to ttf in an other PR

js/packages/components/chat/ChatSettingsMemberDetail.tsx
JohnBerd: `make lint` doesn't work without this line, it appeared on this PR I don't know why

PR #4107, opened 2022-04-28
feat: add eslint config for imports order

JohnBerd:
The React import still have to manually be placed on top of the file


clegirar:
Is not exactly that, the react import will always be in the builtin group of imports but not at the top of it
But i think we don't care if the react import is at the top or not, as long as it's in the builtin group this is the most important 👍

PR #4117, opened 2022-05-03
fix: mini: Trim command spaces after TrimPrefix

jefft0:
In `func (v *groupView) commandParser`, the prefix of a command like "/debug store group_pk metadata" is removed with:

    strings.TrimPrefix(input, prefix+" ")

This is so that the `cmd` parameter for the command only has the args like "group_pk metadata" without the command prefix. But if the user simply enters "/debug store" then the call to `TrimPrefix` does not match. (There is no space after the prefix.) Therefore the `cmd` parameter for the command contains the prefix "/debug store". But `cmd` should be the empty string in this case. This commit changes to call `TrimPrefix` only with the prefix, and to trim the spaces after.

Signed-off-by: Jeff Thompson <jeff@thefirst.org>

### DIFF
--- a/go/cmd/berty/mini/view_group.go
+++ b/go/cmd/berty/mini/view_group.go
@@ -60,7 +60,7 @@ func (v *groupView) commandParser(ctx context.Context, input string) error {
 					return errors.New("not implemented")
 				}
 
-				trimmed := strings.TrimPrefix(input, prefix+" ")
+				trimmed := strings.TrimSpace(strings.TrimPrefix(input, prefix))
 				return attrs.cmd(ctx, v, trimmed)
 			}
 		}


aeddi:
Hey! Thanks for your contribution.

> The DCO check says that the commit message doesn't have Signed-off-by, but it does.

It's strange indeed, I don't know what makes the DCO action fail.
Can you try to amend your commit (without doing any modification) to change its hash and then push force to re-trigger the CI?

aeddi (approved):

jefft0:
> Can you try to amend your commit (without doing any modification) to change its hash and then push force to re-trigger the CI?

I did the amend and push force. No change.

jefft0:
Never mind. I fixed the DCO problem. It wanted me to use my GitHub username, not my full name.

aeddi:
> Never mind. I fixed the DCO problem. It wanted me to use my GitHub username, not my full name.

Ok thanks for the info! 👍

PR #4123, opened 2022-05-06
chore: move utils to new @berty/utils package

n0izn0iz:


clegirar (commented):
What do you think about change naming for:
- utils/sound/soundTypes.ts -> sound.types.ts
- utils/protocol/StreamInProgress.ts -> StreamInProgress.types.ts
- utils/accounts/accountImportExport.ts -> accountBackup.ts

PR #4124, opened 2022-05-06
misc front renaming and removal

n0izn0iz:
Depends on #4123


clegirar (approved):
LGTM 👍 (so good to see firsts refacto PR's 😍)

PR #4138, opened 2022-05-11
fix: refacto front state machine

clegirar:


js/packages/redux/reducers/ui.reducer.ts
```diff
@@ -218,54 +135,41 @@ const slice = createSlice({
...
-		},
-		setStateOpening: setStateOpeningFn,
-		setStateOpeningClients(state: UiState) {
-			state.appState = MESSENGER_APP_STATE.OPENING_WAITING_FOR_CLIENTS
+			Object.keys(initialState).map(k => {
+				if (['accounts'].indexOf(k) === -1) {
+					//@ts-ignore
+					state[k] = initialState[k as keyof UiState]
+				}
+			})
```
n0izn0iz, line 156:
```suggestion
			state.accounts = initialState.accounts
```

js/packages/screens/settings/DevTools/DevTools.tsx
```diff
@@ -576,7 +573,7 @@ const BodyDevTools: React.FC<{}> = withInAppNotification(({ showNotification }:
 			<DropDownPicker
 				items={items}
 				defaultValue={i18next.language}
-				onChangeItem={(item: Item) => dispatch(setAccountLanguage(item.value))}
+				onChangeItem={async (item: Item) => await i18next.changeLanguage(item.value)}
```
n0izn0iz, line 576: no need to await since a promise will be returned already

js/packages/store/messenger-effects/finishPreparingAccount.effect.ts
```diff
@@ -0,0 +1,83 @@
...
+export const finishPreparingAccount = async (
+	ui: UiState,
+	messenger: MessengerState,
+	conversations: Dictionary<beapi.messenger.IConversation>,
+	dispatch: AppDispatch,
+) => {
+	try {
+		await closeConvos(ui.messengerClient, conversations)
+
+		await i18next.changeLanguage(detectOSLanguage())
```
n0izn0iz, line 70: why change language here ?

js/packages/store/messenger-effects/initialLaunch.effect.ts
```diff
@@ -0,0 +1,58 @@
...
+			// Delete berty-backup account
+			const outFile =
+				RNFS.TemporaryDirectoryPath + `/berty-backup-${accountSelected.accountId}` + '.tar'
+			RNFS.unlink(outFile)
+				.then(() => {
+					console.log('File deleted')
+				})
+				.catch(() => {
+					console.log('File berty backup does not exist') // here
+				})
```
n0izn0iz, line 46: why is this needed ?

n0izn0iz (changes requested):
also maybe rename streamError to initError or something

js/packages/redux/reducers/ui.reducer.ts
clegirar: nope it's for not reset accounts

js/packages/store/messenger-effects/finishPreparingAccount.effect.ts
clegirar: yep no needed here, it was done before 👍

clegirar:
streamError is not only set for an initialization error, you can have stream errors when you opening an account but also creating one, deleting one, importing one, updating one, so streamError seems fine like naming no ?

n0izn0iz:
> streamError is not only set for an initialization error, you can have stream errors when you opening an account but also creating one, deleting one, importing one, updating one, so streamError seems fine like naming no ?

hmm, `globalError` maybe ?

clegirar:
> hmm, globalError maybe ?
hmm, for you `streamError` it's only for messenger `EventsStream` ? That's what makes you ?

js/packages/redux/reducers/ui.reducer.ts
n0izn0iz: sorry I read the condition as reverse, I get it now, maybe just do `if (k !== 'accounts')`

n0izn0iz (approved):
LGTM

PR #4142, opened 2022-05-16
feat: atomic button

lucasdc99:


n0izn0iz (approved):
LGTM

clegirar:
For the moment I added a comment to disable eslint for all unused `js/packages/components`, but after the implement of atomic design, we'll keep only used components 👍

PR #4144, opened 2022-05-17
feat: add start/stop method to infra testing tool

gfanton:
add stop/start command to infra testing tools


tool/infra-testing-tool/go/cmd/testflow.go
```diff
@@ -0,0 +1,378 @@
...
+			if err != nil {
+				logging.LogErr(err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func runTestFlow(ctx context.Context, timeout time.Duration, c config.Config) (*Result, error) {
```
aeddi, line 121: I didn't check in detail but wasn't it possible to reuse the `startTest` and `stopTest` functions to avoid duplicating code / be more DRY ?

Something like:
```go
func runTestFlow() {
	startTest()
	handleSignalAndTimeout()
	stopTest()
}
```

tool/infra-testing-tool/go/cmd/testflow.go
```diff
@@ -0,0 +1,378 @@
...
+		}
+
+		logging.Log("waiting for all tests to finish ...")
+
+		// wait for wait group to finish
+		wg.Wait()
+
+	}
+
+	logging.Log("all tests are finished")
```
aeddi, line 375: This log and the log above line 368 are misleading. Something like "waiting for gRPC calls to be sent..." and "all gRPC calls have been sent" should be better.

aeddi: To make it clearer for the user, you could also add an info like: "peers are still running the requested tests... ctrl + c to stop the tests".

tool/infra-testing-tool/Makefile
```diff
@@ -61,6 +61,11 @@ update_ami:
 		-var 'server_go_version=$(SERVER_GO_VERSION)' \
 		.
 
+generate_proto:
```
aeddi, line 64: This rule should depend on the proto files and be a dependency for the build rule I think.

aeddi (approved):
You need to rebase on master to make the Android CI job work.

Other than that and the few comments, LGTM 👍

tool/infra-testing-tool/Makefile
gfanton: totally, but I will use static path for now, since we only have one proto file, I will make something smarter if we get more protos

PR #4145, opened 2022-05-17
feat: Add a CLI to Tyber

D4ryl00:
Add a main CLI


go/pkg/bertymessenger/service.go
```diff
@@ -632,6 +632,12 @@ func (svc *service) sendAccountUserInfo(ctx context.Context, groupPK string) (er
 		})...)
 	}
 
+	// return if there is no user info
+	if acc.GetDisplayName() == "" && avatarCID == "" {
```
n0izn0iz, line 636: we should not return here, the user might want to remove it's display name or avatar

and if we really want this it should return before re-encrypting the avatar imo

D4ryl00: Ok I understand, but currently if there is no display name AND no avatar, the payload in the metadata store is empty, and `handleAppMessageSetUserInfo` returns an error (go/internal/messengerpayloads/handlers.go:941).
What do you advise me?

PR #4160, opened 2022-05-30
fix: js: In README change requirement to Java 15

jefft0:
The latest version of Android Studio uses Gradle 6.7. This is compatible with Java 15, according to:
https://docs.gradle.org/current/userguide/compatibility.html#java

Change the requirement to say Java 15 or earlier.

Signed-off-by: jefft0 <jeff@thefirst.org>


jefft0:
Even though Android Studio can use later versions of the JDK, what matters is support for the runtime on the phone, especially older versions of Android. Better to stay with Java 8. Closing without merge.

PR #4175, opened 2022-06-02
feat: setup e2e testing with appium

n0izn0iz:


js/e2e/README.md
```diff
@@ -0,0 +1,30 @@
...
+npx appium
+```
+
+### Build the app
+
+```sh
+make ios.release
+```
+
+Then grab the app path from the `info Installing` line in the command output, for example:
```
aeddi, line 21: Can you explain here that the user will have to close the simulator at this step if it launches?
That `make ios.release` is just the command used to build the app and launch it, but in the case of the test, we are only interested in the build.

aeddi (approved):
Can you rename the `js/e2e` to `js/e2e-tests`?
Or maybe `js/tests/e2e` if we plan to add later something like `js/tests/UI` with the snapshots for example.

Besides that, LGTM 👍

js/e2e/README.md
n0izn0iz: sure

PR #4181, opened 2022-06-07
feat: ios notification filtering

glouvigny:


js/ios/Entitlements/Prod.entitlements
```diff
@@ -4,6 +4,8 @@
 <dict>
 	<key>aps-environment</key>
 	<string>production</string>
+	<key>com.apple.developer.usernotifications.filtering</key>
```
aeddi, line 7: I'm not sure but maybe if you remove this entitlement from this file and you keep it only in ProdNS.entitlements file, it could be enough.

aeddi: My thought is based on the fact that only the app extension needs to filter if I am not mistaken, and:
- `Prod.entitlements` file is used by the application.
- `ProdNS.entitlements` file is used by the app extension (NS stands for Notification Service).

aeddi (approved):
LGTM 👍

PR #4182, opened 2022-06-08
feat: implementation atomic items

lucasdc99:


js/packages/components/items/floating/FloatingToggleMenuCustomPack.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+
+import { ToggleItemMenuWithIconProps } from '../interfaces'
+import { FloatingToggleMenuPriv } from './FloatingToggleMenu.priv'
+
+export const FloatingToggleMenuCustomPack: React.FC<ToggleItemMenuWithIconProps> = props => {
```
n0izn0iz, line 6: I would prefer to pass PackProps instead of having the *CustomPack variants

js/packages/components/items/floating/FloatingItemMenuWithIconBlue.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+import { useThemeColor } from '@berty/store'
+
+import { ItemMenuWithIconProps } from '../interfaces'
+import { FloatingItemMenuWithIconPriv } from './FloatingItemMenuWithIcon.priv'
+
+export const FloatingItemMenuWithIconBlue: React.FC<ItemMenuWithIconProps> = props => {
```
n0izn0iz, line 8: This variant seems overkill

js/packages/components/items/floating/FloatingItemMenuWithIconBlueCustomPack.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+import { useThemeColor } from '@berty/store'
+
+import { ItemMenuWithIconProps } from '../interfaces'
+import { FloatingItemMenuWithIconPriv } from './FloatingItemMenuWithIcon.priv'
+
+export const FloatingItemMenuWithIconBlueCustomPack: React.FC<ItemMenuWithIconProps> = props => {
```
n0izn0iz, line 8: This variant seems overkill

js/packages/components/items/floating/FloatingItemMenuWithIconBlue.tsx
clegirar: Can we test with an atomic component like that, that just do one thing ?
It's use in many files, and i just want to see if in development, this kind of component, it will be usefull or useless ?

n0izn0iz: if you want but at least call it FloatingMenuItemWithPrimaryIcon
- FloatingItemMenu would mean it's a menu, hence the FloatingMenuItem change
- if the primary color is not blue anymore, we don't need to rename the component

clegirar: Yep the naming it's another thing, but yes i agree 👍 

FloatingMenuItemWithPrimaryIcon seems logic :)

js/packages/components/items/floating/FloatingToggleMenuCustomPack.tsx
clegirar: I agree too for this one, because with the time we'll have only one icon pack, and it's more easily to delete the props than a file, an export, and replace the component in screen 👍

clegirar:
For the naming in general, i think it would be better if we rename ItemMenu, ToggleMenu, FloatingItemMenu into MenuItem, MenuToggle, FloatingMenuItem no ?

clegirar (commented):
For the naming in general, i think it would be better if we rename ItemMenu, ToggleMenu, FloatingItemMenu into MenuItem, MenuToggle, FloatingMenuItem no ?

PR #4201, opened 2022-06-15
feat: initial verifiable credentials issuer

glouvigny:
Suggested test flow

* run mini client `berty mini`
* run a verified credentials server, currently no codes are sent, only phone numbers verified `berty vc-issuer -http.listener 0.0.0.0:7001 -http.server-root http://localhost:7001 -vc-sk JhwhcwAYx7+G7i+XxZJN4ZMaUz9MiYlALlhrd5ilaI4`
* in mini start the verification process `/services credential init http://localhost:7001`
* open the page in a browser (the URL is long, but has been put in the clipboard)
* on the web page, type a phone number with a valid format (ie. `+33123456789`), submit
* on the web page, type the validation code ("000000"), submit
* copy the URI shown on the page, the page includes a redirection to a `berty://vc/` URI
* complete the validation process in mini `/services credential complete {URI}`, mini should display the typed phone number


api/bertyverifiablecreds.proto
```diff
@@ -0,0 +1,71 @@
...
+
+message VerifiableCredsContext {
+  bytes timestamp = 1;
+  string berty_link = 2;
+  CodeStrategy code_strategy = 3;
+  string identifier = 4;
+  string code = 5;
+  string redirect_uri = 6 [(gogoproto.customname) = "RedirectURI"];
+  string state = 7;
+}
```
aeddi, line 34: The name of these methods does not seem ideal. Perhaps evoke something more along the lines of "request/init a challenge" and "respond to/complete the challenge". No specific idea but it's a suggestion.

api/bertyverifiablecreds.proto
```diff
@@ -0,0 +1,71 @@
...
+option go_package = "berty.tech/berty/go/pkg/verifiablecredstypes";
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+
+service VerificationCodeSender {
+  rpc SendVerificationCode (SendVerificationCode.Request) returns (SendVerificationCode.Reply);
+  rpc ValidateIdentifier (ValidateIdentifier.Request) returns (ValidateIdentifier.Reply);
+}
```
aeddi, line 16: This functionality being an integral part of the flow and not intended to be reused should perhaps be integrated as a function rather than split into a separate service.

api/bertyverifiablecreds.proto
```diff
@@ -0,0 +1,71 @@
...
+  message Reply {
+    string sanitized = 1;
+  }
+}
+
+enum FlowType {
+  FlowTypeUndefined = 0;
+  FlowTypeCode = 1;
+  FlowTypeAuth = 2;
+  FlowTypeProof = 3;
```
aeddi, line 63: Can you specify that this mode are not implemented yet and declared only to be future proof?

aeddi (approved):
Besides that, LGTM 👍

PR #4210, opened 2022-06-20
feat: implementation atomic dropdowns 2

lucasdc99:


n0izn0iz (changes requested):
please remove the .idea unless you have an argument for keeping it

PR #4212, opened 2022-06-21
fix: set correct push title on contact convs

n0izn0iz:


D4ryl00 (approved):
LGTM + tested with Android/iOS devices 👍

PR #4238, opened 2022-06-30
feat: fix and use logfilters in release mode

D4ryl00:
You can now use logfilters in devtools even in release mode.

- default logs: `info+:bty*,-*.grpc,error+:*`
- no logs: `-*`
- full logs: `*`


gfanton (changes requested):
It won't work in release mode because there is another protection in `LoggerDriver.swift` that will set log has private message when `CFG_APPSTORE` is set:
see: https://github.com/berty/berty/blob/master/js/ios/Berty/Sources/Bridge/LoggerDriver.swift#L42-L81
```swift
    #if CFG_APPSTORE
    self.scope = Visibility.hidden
    #else
    self.scope = Visibility.visible
    #endif
```
and
``` swift
        switch self.scope {
        case Visibility.visible: os_log("[%{public}@] %{public}@", log: logger, type: type, ulevel, out)
        case Visibility.hidden: os_log("[%{public}@] %{private}@", log: logger, type: type, ulevel, out)
        }
```
you should keep the switch case, but remove defun check

clegirar (changes requested):
And for finish can you rename `js/packages/utils/persistent-options` folder into `js/packages/utils/global-persitent-options` (i know is not your decision for the naming but i think it's more appropriate 👍 )

PR #4261, opened 2022-07-06
feat: implementation member bar in chat

lucasdc99:


js/packages/components/chat/member-bar/MemberBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+				getPeerFromMemberPK({ memberPK: member.publicKey, convPK: props.convId }),
+			)
+
+			if (peer.payload && list.length < 5) {
+				list.push({
+					networkStatus: peer.payload as PeerNetworkStatus,
+					publicKey: member.publicKey ?? undefined,
+				})
+			}
+		})
```
n0izn0iz, line 44: it would be better to shorten the list before calling `forEach` instead of fetching the network status for the whole list and only pushing 5

I guess you do that to still have 5 elements in case some peer.payload is undefined?

In this case maybe use a `for (const member of members)`and break out of the loop after the list if full

n0izn0iz (approved):
LGTM except comment

js/packages/components/chat/member-bar/MemberBar.tsx
clegirar: Nope we still 5 elements because the UI show only 5 peers in the bar but in this case too i think too that it's efficient to use `for (const member of members)` and break out of the loop when we have 5 elements

PR #4274, opened 2022-07-12
fix: bottom sheet modal android crash, replace BottomSheet with Modal of react-native

clegirar:


D4ryl00 (approved):
Tested on Android

PR #4316, opened 2022-07-25
fix: refacto permissions screens

clegirar:


js/packages/components/chat/footer/recorder/RecorderButton.tsx
```diff
@@ -145,21 +128,30 @@ export const RecorderButton: React.FC<RecorderButtonProps> = ({
...
-		setCurrentTime(Date.now())
-		setClearRecordingInterval(setInterval(() => setCurrentTime(Date.now()), 100))
-		meteredValuesRef.current = []
+		await checkPermission({
+			permissionType: PermissionType.audio,
+			navigate,
+			accept: () => {
+				// We check the status before ask it, to don't record after asking permission
+				// (In other apps, the recorder don't start after accept permission)
+				if (status === RESULTS.GRANTED) {
```
n0izn0iz, line 139: I think `status` is out of date here

js/packages/components/chat/footer/CameraButton.tsx
```diff
@@ -1,104 +1,29 @@
...
-
 	return (
 		<ChatInputButton
 			iconName='camera'
 			iconPack='custom'
 			vOffset={1.5}
 			iconRatio={0.44}
-			onPress={handlePressCamera}
+			onPress={async () => {
+				await handlePressCamera({ sending, setSending, messengerClient, onClose, navigate })
```
n0izn0iz, line 25: no need to use async/await here I think

js/packages/utils/permissions/handle-press-permissions/prepareMediaAndSend.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const prepareMediaAndSend: (props: PrepareMediaAndSendProps) => void = async ({
+	sending,
+	setSending,
+	messengerClient,
+	onClose,
+	res,
+}) => {
+	try {
+		if (sending) {
```
n0izn0iz, line 16: this might be out of date

n0izn0iz (approved):
LGTM except comments

js/packages/components/chat/footer/CameraButton.tsx
clegirar: Yes we need, in `handlePressCamera` function we have `checkPermission` await function that is necessary to await 👍

js/packages/components/chat/footer/recorder/RecorderButton.tsx
clegirar: What do you mean by out of date ?

js/packages/utils/permissions/handle-press-permissions/prepareMediaAndSend.ts
clegirar: What do you mean by out of date ?

js/packages/components/chat/footer/CameraButton.tsx
n0izn0iz: using `() => handlePressCamera({ sending, setSending, messengerClient, onClose, navigate })` will return a promie and is the same as wrapping it in async/await

js/packages/components/chat/footer/recorder/RecorderButton.tsx
n0izn0iz: I mean that by the time the `accept` callback is called, the status value might be different

js/packages/utils/permissions/handle-press-permissions/prepareMediaAndSend.ts
n0izn0iz: for example here: https://github.com/berty/berty/pull/4316/files/d4a1108ee45312079635f76887c02a6670915ae9#diff-a4fa0c07ae99f80ca0174f3eed339c9d3abb1588e397e5d012c7a517d5666aa1R32

by the time this part is run, the value of sending might be different

js/packages/components/chat/footer/recorder/RecorderButton.tsx
clegirar: Mhhh, but this value is usefull just at the init

clegirar: But i want the `status` value before call `accept` callback

clegirar: I can change his name to be more clear like `initialStatus` something like that

PR #4347, opened 2022-08-08
wip: berty bot experiments

moul:


go/pkg/bertybot/recipes.go
```diff
@@ -72,7 +72,6 @@ func WelcomeMessageRecipe(text string) Recipe {
 			}
 		},
 	}
-	// FIXME: send welcome message to new conversations
```
Doozers, line 75: this has been fixed automatically

PR #4357, opened 2022-08-11
chore: move bertylinks & dependencies from internal to pkg

Mikatech:
# Hello, :wave: 

I am currently working on **implementing a berty bot for an application.** I need my bot to **be able to use the functions of the Bertylinks** module which is currently in internal  in order to manage group invitations.

## modification :factory_worker: 

To solve this problem, I propose to **move the Bertylinks module in pkg**, moreover this module uses cryptoutil, so I move the file necessary to Bertylinks in a new module crypoutil in pkg.

## Verification :white_check_mark: 

I have performed the tests which do not show any difference with the result obtained with the current state
I also performed functional tests of berty using berty-mini and my mobile device.


moul:
👋 Thank you for your contribution.

I agree w/ the need to move `bertylinks` to `./pkg/`.
I was wondering why we also need to move the `cryptoutil`.

Can you provide a quick snippet of the expected usage from outside Berty?

Mikatech:
Cryptoutil was used in bertylinks and at first I thought it had to be in pkg to be used in it but it seems not.

An error of rigor on my part on this, I will fix it ASAP !

Thanks for the review @moul

Mikatech:
> wave Thank you for your contribution.
> 
> I agree w/ the need to move `bertylinks` to `./pkg/`. I was wondering why we also need to move the `cryptoutil`.
> 
> Can you provide a quick snippet of the expected usage from outside Berty?

That was an error, i fixed it !
Thanks :star:

moul (approved):

PR #4374, opened 2022-08-21
chore: configure GitHub Actions’ concurrency

moul:
Should be better than current ways of handling concurrency.

If it works well, I suggest to port it on every other relevant workflow.


jefft0:
We don't have the resources to test this right now, but it could be useful. Putting in the backlog.

PR #4380, opened 2022-08-23
feat: add tests for all chat files

lucasdc99:
Fix https://github.com/berty/zenhub/issues/5


clegirar:
We have deleted problematic screens (OneToOne and MultiMember, we have diff between master and local), to merge the rest of the screens and we'll dig this screens in an other PR 👍

PR #4400, opened 2022-08-31
fix: delete Bertypush.xcframework in js Makefile clean rule

clegirar:


aeddi (approved):
LGTM 👍

PR #4462, opened 2022-10-11
chore: move multi account features in devtools

clegirar:
Screenshot 2022-11-17 at 10 52 09
Screenshot 2022-11-17 at 10 52 29


js/packages/i18n/locale/en-US/messages.json
```diff
@@ -126,10 +126,6 @@
 			"search": {
 				"no-results": "No results found"
 			},
-			"multi-account": {
-				"create-button": "Create new account",
-				"import-button": "Import account"
-			},
```
glouvigny, line 132: Maybe we should avoid moving keys around as it will likely break translations in other languages

clegirar: these specific keys are no longer used, but i'll avoid moving used keys 👍

glouvigny (approved):

PR #4472, opened 2022-10-20
feat: select a remote node from mobile app

D4ryl00:
- [x] add new methods in native frameworks and BertyBridge to init the bridge with a GRPC client trying to connect to a remote node
- [x] add a new screen when the app starts to select the internal or remote node. Add the ability to change values from Devtools

closes https://github.com/berty/zenhub/issues/40


gfanton (approved):
LGTM 👍

PR #4538, opened 2023-01-03
feat: disable public dht

gfanton:


aeddi-sudo:
Is it not possible to also disable LAN DHT since we don't use it either? (Peer discovery w/ mDNS, Group rdv discovery w/ OrbitDB / Pubsub and Contact Request rdv discovery w/ Tinder local driver)

aeddi (approved):
LGTM 👍

PR #4590, opened 2023-02-16
fix: activateGroupContext: wait for the watch new members goroutine to add the device key

jefft0:
`activateGroupContext` calls `activateGroupContextactivateGroupContext` and starts a goroutine to monitor the channel for public keys of added devices. If `selfAnnouncement` is true, then `AddDeviceToGroup` is called so that the goroutine should receive the member's own device key. `activateGroupContext` can return before the asynchronous goroutines are signalled to add the device keys. Another function like `ActivateGroup` which calls `activateGroupContext` can return before all the needed keys are added. This is a bug because `ActivateGroup` does not finish activating the group because the needed keys are not added.

This pull request updates `activateGroupContext` to add a `WaitGroup` which is signalled by the `WatchNewMembersAndSendSecrets` goroutine. If `selfAnnouncement` is true and `AddDeviceToGroup` does not return nil (where nil means the device was already added) then wait for the `WaitGroup` to signal that the key was added. Now, when `activateGroupContext` returns, it means that the group context is fully activated.

This fix was tested on an Ubuntu 22.04 virtual machine (where the flappy tests fail more often then on macOS). This fix seems to make the following flappy tests pass:

    TestFlappyRestoreAccount
    TestRaceReactivateAccountGroup

### DIFF
--- a/go/pkg/bertyprotocol/group_context.go
+++ b/go/pkg/bertyprotocol/group_context.go
@@ -104,6 +104,9 @@ func (gc *GroupContext) activateGroupContext(contact crypto.PubKey, selfAnnounce
 	// syncChMKH := make(chan bool, 1)
 	// syncChSecrets := make(chan bool, 1)
 
+	wgFinished := sync.WaitGroup{}
+	wgFinished.Add(1)
+
 	{
 		// Fill keystore
 		chNewData := gc.FillMessageKeysHolderUsingNewData()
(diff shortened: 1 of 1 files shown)


jefft0:
@gfanton , after splitting the code into weshnet, maybe we should close the pull request and use https://github.com/berty/weshnet/pull/9 .

jefft0:
Close in favor of https://github.com/berty/weshnet/pull/9 .

PR #4608, opened 2023-02-23
chore: Link to the primary copy of the Wesh protocol doc

jefft0:
To avoid multiple copies of the same document which can diverge, point to the primary Wesh protocol doc at https://berty.tech/docs/protocol . We can remove the img directory which is duplicated there (and may diverge).


D4ryl00 (approved):
LGTM

PR #4616, opened 2023-02-28
feat: devtools, search contacts from user's address book

glouvigny:


clegirar (approved):
LGTM for the js part and testing 👍

sudwhiwdh:
@glouvigny Read phone numbers, email addresses, Android contacts. I am irritated. What does this have to do with Berty? Why does a data-saving app that does not use classic identifiers need/implement this function?

glouvigny:
> @glouvigny Read phone numbers, email addresses, Android contacts. I am irritated. What does this have to do with Berty? Why does a data-saving app that does not use classic identifiers need/implement this function?

This is an optional feature aimed at less tech-savvy or less privacy conscious people, having it listed in the Android app manifest does not mean the app will read your contacts.

PR #4620, opened 2023-03-02
make Makefile asdf target POSIX sh compatible

asmodehn:
Ref: https://stackoverflow.com/a/3411105

Small change I had to make to get the Makefile commands work on my machine (`/bin/sh` is symlinked to dash on my PopOS).
This should still work the same with bash.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ asdf.install_plugins:
 	@echo "Installing asdf plugins..."
 	@set -e; \
 	for PLUGIN in $$(cut -d' ' -f1 .tool-versions | grep "^[^\#]"); do \
-		asdf plugin add $$PLUGIN || [ $$? == 2 ] || exit 1; \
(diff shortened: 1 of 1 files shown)


jefft0:
Hello @asmodehn . We had a similar issue in a related project. I think we want to keep the double `==`. We fixed it with this pull request. https://github.com/berty/weshnet/pull/18
On your system, can you try `$$?==2` (without spaces)?

asmodehn:
Indeed that works too :+1:

jefft0:
Glad it works for you! Can you squash into one commit, and add a "Signed-off-by" according to the DCO test? https://github.com/berty/berty/pull/4620/checks?check_run_id=11751451946

asmodehn:
Done. all looks good now.

gfanton (approved):
LGTM 👍

PR #4636, opened 2023-03-20
fix: update weshnet

gfanton:
depends on: https://github.com/berty/weshnet/pull/39

should correctly update dependencies with wesh new update


D4ryl00 (approved):
LGTM

Issue #4626, opened 2023-03-10
unmute popup is very dark


LFGaming:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Berty product

Mobile app

### Berty product version

n&#x2F;a (n and hastag x2F semicolon a)

### OS

Android

### OS version

Android version 13

### Device

Samsung Galaxy S20 5G

### Steps to reproduce

1. join a group
2. go to the normal settings (there where you can activate notifications and bluethooth etc)
3. click the bluetooth option
4. click a group at the botom to unmute
5. the popup will be very dark

### Current behavior

The popup to unmute a group is very dark see picture:

![photo_2023-03-10_18-34-21](https://user-images.githubusercontent.com/38732192/224384670-19778627-90d7-42c2-8b6e-0329ede0800d.jpg)

### Expected behavior

should be less dark

PR #4639, opened 2023-03-23
fix: remove Android dark theme

iuricmp:
- App was using the Android dark theme as default.

Reference: https://developer.android.com/develop/ui/views/theming/darktheme

Fix https://github.com/berty/berty/issues/4626

### DIFF
--- a/js/android/app/src/main/res/values/styles.xml
+++ b/js/android/app/src/main/res/values/styles.xml
@@ -1,7 +1,7 @@
 <resources>
 
     <!-- Base application theme. -->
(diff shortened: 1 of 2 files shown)


clegirar (commented):
Instead of change `DayNight` by `Light`, is it possible to follow the theme system setting ?

sudwhiwdh:
> Instead of change DayNight by Light, is it possible to follow the theme system setting ?

Cannot test this unfortunately without F-Droid version: Are there already the options for choosing day, night and to follow the system theme style? In principle, I would find all three options important to have. And as a user, I would then also prefer the option "follow the system theme style" as the default.

clegirar (approved):
LGTM 👍

PR #4646, opened 2023-03-29
fix(ui): small label position fix & more stories for storybook

iuricmp:
- small fix on label position

Screenshot 2023-03-29 at 09 43 29

- add onboarding stories

https://user-images.githubusercontent.com/689440/228478398-238a8e34-465a-4f10-bbd1-064728faff50.mp4

- i18n support

i18n

### DIFF
--- a/js/.storybook/addons.tsx
+++ b/js/.storybook/addons.tsx
@@ -1,0 +1,29 @@
+import React from 'react'
+import addons from '@storybook/addons'
+import i18next from 'i18next'
+import { languages } from '@berty/i18n/locale/languages'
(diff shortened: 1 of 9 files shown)


js/package.json
```diff
@@ -7,6 +7,7 @@
     "ios": "react-native run-ios",
     "start": "react-native start",
     "test": "jest --silent",
+    "teste-snapshot-update": "yarn test -u",
```
clegirar, line 10: I think you can add a rule in the Makefile, instead of a script in the `package.json`

clegirar (approved):
LGTM 👍

Issue #4663, opened 2023-04-21
Berty app does not compile with Xcode 14.x


iuricmp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Berty product

Mobile app

### Berty product version

2.466

### OS

iOS

### OS version

13.0

### Device

iPhone 14

### Steps to reproduce

1. $ cd berty/js
2. $ make ios.debug

### Current behavior

![image](https://user-images.githubusercontent.com/689440/233624662-c4a17ffa-1a1a-44c8-bae8-63dfda56d5f9.png)

### Other

Related to https://stackoverflow.com/questions/75897834/use-of-bitwise-with-boolean-operands-xcode-14-3-fails-builds-using-react-n

PR #4664, opened 2023-04-21
fix: patch-package Yoga to avoid issues on RN 0.64.4 builds with XCode 14.x

iuricmp:
The app build fails when building with the latest XCode version.

Fix: https://github.com/berty/berty/issues/4663


D4ryl00 (approved):
Tested on Xcode 14.3.

PR #4666, opened 2023-04-27
fix: CI Runners with XCode 14.3

D4ryl00:
When we run e2e CI tests, we try to run them with an iPhone 11 but with XCode 14, only iPhone 14 simulators are available by default.


aeddi (approved):
LGTM 👍

PR #4667, opened 2023-04-28
fix: Delete deep links handler from account's screens

iuricmp:
The direct link was not being entered in the chat screen.
This PR deletes the deep links handler from the account's screens because the messenger client is not initialized yet.

### DIFF
--- a/js/packages/navigation/stacks.tsx
+++ b/js/packages/navigation/stacks.tsx
@@ -135,7 +135,9 @@ Components = mapValues(RawComponents, SubComponents =>
 		(Component: React.FC): React.FC =>
 			React.memo(props => (
 				<>
-					{Platform.OS !== 'web' ? <DeepLinkBridge /> : null}
+					{Platform.OS !== 'web' && !('OpeningAccount' in SubComponents) ? (
+						<DeepLinkBridge />
+					) : null}
 					<Component {...props} />
 				</>
 			)),


js/packages/navigation/stacks.tsx
```diff
@@ -126,22 +126,35 @@ const DeepLinkBridge: React.FC = React.memo(() => {
 	return null
 })
 
-let Components: typeof RawComponents
+let Components: any
```
clegirar, line 129: If you can found a type for this instead of `any`, it could be cool 👍

clegirar (changes requested):

clegirar:
And it's not really an inject of deeplink into Chat's screens, it's more delete deeplink for Account's screens because the messenger client is not initialized yet 👍

js/packages/navigation/stacks.tsx
```diff
@@ -126,22 +126,40 @@ const DeepLinkBridge: React.FC = React.memo(() => {
 	return null
 })
 
-let Components: typeof RawComponents
+let Components: {
+	Chat: typeof Chat
+	Settings: typeof Settings
+	Onboarding: typeof Onboarding
+	Account: typeof Account | any
```
clegirar, line 133: Why `| any` ?
Please if you can find a way without `any` types 👍 Just `typeof Account` doesn't works ?

clegirar (changes requested):

js/packages/navigation/stacks.tsx
```diff
@@ -130,16 +130,15 @@ let Components: typeof RawComponents
...
-		(Component: React.FC): React.FC =>
-			React.memo(props => (
-				<>
-					{Platform.OS !== 'web' ? <DeepLinkBridge /> : null}
-					<Component {...props} />
-				</>
-			)),
-	),
+	mapValues(SubComponents, (Component: React.FC): React.FC => {
+		console.log('SubComponents', SubComponents)
```
clegirar, line 134: Please remove the `console.log` 👍

clegirar (changes requested):

clegirar (approved):
You can just directly return the `React.memo` instead of `return React.memo`, but it's just a little thing 👍 else LGTM 👍

PR #4678, opened 2023-05-12
Mini readme

LFGaming:
Added a little explenation on how to use berty mini to the CLI readme, so people understand how the basics work to play around.


jefft0:
Please fix DCO check and squash into one commit.

PR #4689, opened 2023-05-31
In .tool-versions comments, remove version numbers and dates in comments

jefft0:
Remove specific version numbers and dates in the comments because they get out of sync when we do global updates of actual version numbers. Put a comment at the top of the file explaining that the "as of" dates can be seen in git blame.


.tool-versions
```diff
@@ -1,11 +1,14 @@
+# To see the date when a version was updated, use git blame:
+# https://github.com/berty/berty/blame/master/.tool-versions
+
 #-----
-# golang 1.18.4 is simply the most recent version available to date (07/22).
+# This is simply the most recent Go 1.19 version available to date.
```
D4ryl00, line 5: There is still the version of Go here, is it what you want?

D4ryl00 (approved):
LGTM

.tool-versions
jefft0: We are not using the "latest to date" version of Go. So in this case, I think we need to explain that we are keeping the version at 1.19. Alternatively, we could remove the version and say "This is simply the most recent version of the lowest major version of Go which is allowed by kubo."  This is the truth, but it's a little complicated. What do you think?

D4ryl00: I think we can be a little complicated here as we are targeting dev users.

jefft0: OK. I changed to the more complicated comment. Merging.

PR #4691, opened 2023-06-06
fix: duplicated member entry

### LINKED ISSUE #4687: anon user gets added to every group made

### Berty product

Mobile app

### Berty product version

Android version v2.268.2.dev.2, iOS version: 2.468.3

### OS

iOS

### OS version

Android version 12 and iOS version 16.4.1

### Device

iPhone 12 mini and Samsung Galaxy S20 5G

### Steps to reproduce

1. Make a group
2. Open the group
3. Click the group settings in the top right
4. Click members
5. There is a anon#{Identifier} in the group

### Current behavior

There is a anon# user always added to every group even when you did not add them, they are also always offline.

See Image:
photo_2023-05-30_21-23-55

### Expected behavior

I would expect there would not be any extra users added in the group?

### Other

_No response_

iuricmp:
Fix for duplicate member in group members list.

Disclaimer: This is a quick-fix solution. The proper solution came from understanding why `TypeMemberUpdated` event is being dispatched twice for the current user in the StreamEvent.

It is a Berty's Javascript issue since it's not reproducible on Berty's Mini.

fix #4687

### DIFF
--- a/js/packages/redux/reducers/messenger.reducer.ts
+++ b/js/packages/redux/reducers/messenger.reducer.ts
@@ -195,20 +195,6 @@ const makeRoot = <T>(val: T) => ({
 const rootInitialState = makeRoot(initialState)
 type LocalRootState = typeof rootInitialState
 
-/**
- *
- * Actions
- *
- */
-
-const defaultDisplayName = (seed: string | null | undefined) => {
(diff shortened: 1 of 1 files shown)


D4ryl00 (approved):
LGTM

PR #4697, opened 2023-06-12
chore: Don't run workflow jobs on forks

jefft0:
PR https://github.com/berty/weshnet/pull/61 was successful to prevent running the codeql workflow jobs when syncing a fork. This pull request does the same for the workflow jobs in this repository. Only run the job if github.repository_owner == 'berty' .

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -24,6 +24,8 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


jefft0:
GitHub now allows the user to disable workflow actions in the fork. This PR is not needed. Closing.

PR #4700, opened 2023-06-12
Fix: only subscribe messages once

jefft0:
weshnet version v1.10.0 has PR https://github.com/berty/weshnet/pull/71 which fixes the message queue to run in a separate goroutine and not block the main event handler. In this PR, we update berty to use this version and we don't need to use a goroutine when calling HandleAppMessage . Finally, we update ActivateGroup to check if the group is already activated and in this case we don't call subscribeToGroup again. This means there is only one subscription which is calling HandleAppMessage.

This fixes the bug where the app freezes when receiving up to 75 groups of 40 messages from Berty mini.


gfanton (approved):
LGTM 👍

Issue #4674, opened 2023-05-05
App gets stuck on loading screen


LFGaming:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Berty product

Mobile app

### Berty product version

V2.468.0

### OS

iOS

### OS version

iOS 16.4.1 (a)

### Device

iPhone 12 mini

### Steps to reproduce

1. Remove app
2. Download version V2.468.0 (latest at this point)
3. Make an account
4. After the making you will be stuck at the three dots that bounce up and down

### Current behavior

Currently if you update to version V2.468.0 the app will freeze at the `setup-local-messenger-server` screen. See Image:
![Screenshot_20230505_140507_Berty_Yolo](https://user-images.githubusercontent.com/38732192/236457327-869c08f9-c1d1-4004-a85b-1830f717db7a.jpg)

If you delete the app and redownload the following will happen:
iOS:
You can make an account but after finishing your account making you will get stuck at the three dots screen. See image:
![IMG_7758](https://user-images.githubusercontent.com/38732192/236457811-89ce01dd-659e-4e0d-8a25-c2fa364a2945.png)

Android:
Some times the app freezes at the following screen:
![Screenshot_20230505_141017_Berty_Yolo](https://user-images.githubusercontent.com/38732192/236457979-45c62e59-c19b-49dc-872c-ad51eebf063e.jpg)
This does not happen everytime after removing and reinstalling the app so it might be something else

### Expected behavior

It should not freeze.

LFGaming:
> I have also seen Berty get stuck at "setup-local-messenger-server".

For me it seems to be very in consistent, sometimes when redownloading the app it works and sometimes it freezes

LFGaming:
Now in version 2.468.3 it freezes always when starting the app for me. but it seemed to be fixed in between the versions.
Anyone that can confirm that it always freezes in version 2.468.3?

jefft0:
@LFGaming : Yes I just tried restarting 2.468.3 on my iPhone 14 and it is stuck at "setup-local-messenger-server".

LFGaming:
After a bit of trying around I found the following:
When using two phones, in my tests they where a iPhone 12 mini running iOS 16.4.1 and an Samsung Galaxy S20 5G running Android 12. With Berty versions: Android version v2.268.2.dev.2, iOS version: 2.468.3

When I had my iphone on and made a chat with the Android that was online everything went fine.
[...]

PR #4703, opened 2023-06-19
Fix: gorm migration error

D4ryl00:
fixes https://github.com/berty/berty/issues/4674

Our gorm sqlcipher driver is outdated and didn't populate correctly the gorm.ColumnType structure. I updated it based on the official sqlite driver.
Our dependency https://github.com/mutecomm/go-sqlcipher is outdated and doesn't support the use of `uniqueIndex`.


gfanton (approved):
LGTM 👍

PR #4715, opened 2023-06-28
Use weshnet v1.12.1 , update API event message names

jefft0:
weshnet v1.12.1 has a breaking change for the names of API event messages. See https://github.com/berty/weshnet/issues/79 for the rationale. Did `make regenerate` .


jefft0:
@D4ryl00, I did `make regenerate` in the ./js folder. Do I need to do anything else after the udpate?

D4ryl00 (approved):
LGTM

PR #4716, opened 2023-06-29
Avoiding Anonymous users from User List

iuricmp:
It hides the anon users from users list but it preserves Berty mini users at the same time.


jefft0:
This uses the front end to fix a problem which may come from the back end. We should first check if we can fix it in the back end. Set to Draft while we check.

Issue #4733, opened 2023-07-19
App crashes when turning Bluetooth mode on


LFGaming:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Berty product

Mobile app

### Berty product version

V2.468.13-dev.10+…

### OS

Android

### OS version

Android 13

### Device

Samsung galaxy S20 5G

### Steps to reproduce

1. Open the app
2. Go to settings
3. Enable Bluetooth
4. Go back to Home Screen (the app will reload)
5. While reloading it crashes

also happens on other android phones (os versions of these phones I do not know)

### Current behavior

The app crashes when you enable Bluetooth mode on android.

https://github.com/berty/berty/assets/38732192/78fd39cd-24f9-4864-b7fb-448386c63a81

### Expected behavior

It should not crash

### Other

This bug came into existence after the update to Android API 31.

leonderdunna:
also happens on Pixel 7 Pro with Android 13

jefft0:
Re-open this bug until the fix is release on the app stores.

PR #4736, opened 2023-07-24
fix: Blueetooth permissions for Android devices and above

D4ryl00:
Since Android 12, no more needs to ask for `ACCESS_FINE_LOCATION` to enable Bluetooth. Instead, we have to request 3 different Bluetooth permissions to the user: `BLUETOOTH_SCAN`, `BLUETOOTH_ADVERTISE`, and `BLUETOOTH_CONNECT`.
We have also to support Android API level 31 and above that introduces these new permissions.

fixes https://github.com/berty/berty/issues/4733

Tested on iPhone 7 and 8.
Tested on Android 13 and 10.


iuricmp (approved):
:shipit:

PR #4740, opened 2023-08-01
feat: disabling the "Submit a bug" button for now

iuricmp:
Disabling it until we have time to improve this feature.


jefft0 (approved):
Works for me in the emulator.

iuricmp:
FYI: Lint was complaining about the unused translation string. So I removed them.

PR #4794, opened 2024-01-19
chore: Fix text input background color and add additional help command

### LINKED ISSUE #4390: Create group: wrong background color

### Berty product

Mobile app

### Berty product version

2.431.1

### OS

iOS

### OS version

15.6.1

### Device

iPhone 13 Pro Max

### Steps to reproduce

1. Login to the Berty app 
2. Click on the + button at bottom right corner 
3. Click on the create group button

### Current behavior

Add member background is yellowish white. It is completely diffrent from any color that we have in application. If you move to the next screen, you will notice that Group info has same issue

image

### Expected behavior

This section should be white(#fff) and y search placeholder should be gray. For reference see color pallet of Inbox screen

### Other

_No response_

iuricmp:
This pull request fixes the background color of the text input to match the Figma design. 
The screen background has already been fixed in another PR.

It also adds an additional yarn command for regenerating the snapshot files.

Fix #4390

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -7,6 +7,7 @@
     "ios": "react-native run-ios",
     "start": "react-native start",
     "test": "jest --silent",
+    "test-snapshot-update": "jest -u",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
     "postinstall": "patch-package && node ./fix-types.js && npx jetify",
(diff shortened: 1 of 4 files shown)


jefft0 (approved):
Tested on Android and iOS simulators. Colors look good.

PR #4637, opened 2023-03-22
WIP chore: move push notifications into berty

D4ryl00:

PR #4795, opened 2024-01-19
chore: update Android API version 33 + RN dependencies

D4ryl00:
This PR updates the Android API version to the 33rd to be able to push it on the Play Store.
We also had to update two React-Native dependencies to be able to compile Berty:
- https://github.com/react-native-cameraroll/react-native-cameraroll/releases/tag/v5.0.1
- https://github.com/react-native-webview/react-native-webview

I also had to fix Android Studio which has a bug with API 33:
https://stackoverflow.com/questions/77172677/android-studio-error-installed-build-tools-revision-33-0-1-is-corrupted-in-rea


D4ryl00:
Already done in PR https://github.com/berty/berty/pull/4637

PR #4796, opened 2024-01-19
Fix translation escaping unicode chars

### LINKED ISSUE #4371: Text draft shows &#x2F for a /

### Berty product

Mobile app

### Berty product version

2.429.3

### OS

iOS

### OS version

ios 15.5

### Device

iPhone 12 mini

### Steps to reproduce

1. Go to a chat,
2. Type `/` but not yet send it,
3. Go back to your screen where you can choose the chats,
4. It will show `Draft &#x2F` instead of a /

### Current behavior

It will show draft &#x2F instead of the /

### Expected behavior

It should show the /

### Other

_No response_

iuricmp:
This pull request fixes an issue with translation escaping unicode characters. I've added the "interpolation: { escapeValue: false }" to avoid escaping the unicode.

fix #4371

### DIFF
--- a/js/packages/screens/chat/Home/components/conversation/ConversationItem.tsx
+++ b/js/packages/screens/chat/Home/components/conversation/ConversationItem.tsx
@@ -72,6 +72,7 @@ export const ConversationItem: React.FC<
 	} else if (chatInputText) {
 		description = t('main.home.conversations.draft', {
 			message: chatInputText,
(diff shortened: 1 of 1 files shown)


jefft0 (approved):
Tested on Android an iOS emulators. The correct character is shown.
Approve, pending DCO filx.

Issue #4415, opened 2022-09-09
Info icons in Custom Mode lead nowhere


ahstro:
### Is there an existing issue for this?
* [x] I have searched the existing issues

### Berty product
Mobile app

### Berty product version
‎2.431.2‎‏‎‎‏‏‏‎‎‏‎‎‏‎

### OS
Android

### OS version
Lineage OS 18.1 (Android 11)

### Steps to reproduce
1. Open fresh Berty install
1. "Create an account"
1. "Custom Mode"
1. "Proceed"
1. Press lil' "circled 'i'"/info symbol
1. Get no smarter but a little wiser 🧙

### Current behavior
![Screenshot!UNITO-UNDERSCORE!20220909-155056!UNITO-UNDERSCORE!Berty](https://user-images.githubusercontent.com/3090888/189367561-449dd3a4-aed9-41cd-ba94-d40810dcf5d3.png)

### Expected behavior
Show some info about the related section :information_source:

### Other
Cool project y'all, hope I can trust it :black_heart::heart:

peletron:
@ahstro thanks for reporting the issue. Our dev team will take a look. 

p.s We happy that you like project :)

jefft0:
@iuricmp to check if there is a straightforward fix.

PR #4797, opened 2024-01-26
Fix info icons in Custom Mode

iuricmp:
This pull request fixes the issue with the info icons in Custom Mode that were not leading anywhere. 
The changes include hiding an unused button and adjusting the alignment and padding of the info icon. This resolves issue #4415.


jefft0 (approved):
Builds and runs on Android an iOS simulators. Custom screen looks good.

PR #4798, opened 2024-01-26
Fix missing translation

iuricmp:
This pull request fixes a missing translation in the `messages.json` file. The translation for the `delete-account` section was incomplete, and this PR adds the missing translation.


jefft0 (approved):
Tested on Android and iOS simulators. The delete message looks good.

Issue #4829, opened 2024-04-05
Fix CI for docker-compose


jefft0:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Berty product

Mobile app

### Berty product version

latest

### OS

iOS

### OS version

Linux CI

### Steps to reproduce

Make a PR. Look at the CI report.

### Current behavior

The CI gives the error:
```
Run docker-compose build
/home/runner/work/_temp/9ea67a76-a6da-[4](https://github.com/berty/berty/actions/runs/8537024699/job/23386837424#step:3:5)9af-9926-9edc0ad81a5f.sh: line 1: docker-compose: command not found
Error: Process completed with exit code 127.
```

### Expected behavior

Should use `docker compose`.

PR #4830, opened 2024-04-05
fix: Change docker-compose to docker compose to fix CI

jefft0:
Fixes issue https://github.com/berty/berty/issues/4829


D4ryl00 (approved):
Please view my comment, otherwise LGTM

PR #4837, opened 2024-05-03
fix: In config.yml, update addresses for rdvp and static-relays

jefft0:
Tested with Android phone on cell network and iOS simulator on house network.


D4ryl00 (approved):
LGTM

PR #4844, opened 2024-05-31
refactor: fix web tests

iuricmp:
This pull request includes a fix for the tests on the web project. It resolves an issue that was causing the tests to fail.

I'm using "refactor" as commit message to avoid releasing a new app version.


jefft0:
@jefft0 to do `make test`.

PR #4853, opened 2024-07-29
chore: adding missing library to web project

iuricmp:
Fix the `make` default command on root folder that generates the following issue:

Screenshot 2024-07-29 at 11 28 29


jefft0 (approved):
Tested on my macOS with `make`. There is no more error.

PR #4866, opened 2024-09-20
Chore/update kubo@v0.29.0 + weshnet

D4ryl00:
This PR continues the PR series across the dependency stack to update kubo to v0.29.0 and libp2p to v0.34.1.

Some other fixes have been made:

- Berty Messenger uses now the v2 of weshnet (berty.tech/weshnet/v2).
- Go version is updated to v1.22.5.
- Some deprecated flags are removed from Goreleaser CI jobs.
- Buf is updated and the configuration files are updated to V2.
- gogoproto has been removed, so we updated the proto files to use a new plugin to tag fields for gorm. Because gogoproto also customized the field names, we are now using standard name camel case convention (accountID -> accountId). Enums have now prefix.
- Quic maddr should now be `quic-v1` instead of `quic`. Config files and the code base has been updated accordingly.
- The new golangci-lint version is more accurate, so we fixed many unused variables/parameters. Objects containing mutex are now passed by a pointer instead of directly in the event system.

There are some issue that will be fixed in future PR :
- Android build error that cannot download gradle-2.3.3 (removed from remote servers). We must update our dependencies.


jefft0 (approved):
Tested on macOS between berty-mini and the iOS simulator. (Will test Android after a later PR to update gradle.)

PR #4870, opened 2024-10-07
chore: bump-golang-1.23.2

D4ryl00:
This PR updates golang to version 1.23.2 that fixes an issue where running the binary `berty` fails (either by doing `go build` or `go install`).


jefft0 (approved):
Tested Berty Messenger in iOS simulator and berty-mini on macOS Sonoma 14.6.1 and Sequoia 15.0.1.

PR #4878, opened 2024-10-17
chore: expose mdns package

D4ryl00:
The mDNS package has been moved from `go/internal` to `go/pkg` to be exposed and used by an external projects (like `weshnet-expo`).

Note: golangci-lint seems blocked when running in Github Action. After trying with the latest version (1.61.1), it seems better. But this version shows new errors that we should fix. To fix one of these, we need to update the `echo` API, which needs the latest weshnet API version.


jefft0 (approved):
Tested building for iOS simulator.

jefft0:
Latest commits work with `make ios.app_deps` .

PR #4895, opened 2024-12-13
chore: .tool-versions: Update to ruby 3.2.2

jefft0:
Update to ruby 3.2.2 (same version as gnonative, etc.) Hopefully, this will fix the CI build on macOS 15. On my local macOS, `make tests` passes and Berty Messenger builds and runs on iOS simulator.


D4ryl00 (approved):
LGTM

PR #4922, opened 2025-03-28
fix: upgrading ci cache

iuricmp:
This PR is an attempt to solve the issue on Berty's CI/CD [reference](https://github.com/berty/berty/actions/runs/13824623500/job/38677218211?pr=4916)
`
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2.1.6`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
`

This PR also solves issues from the latest version of ASDF (0.16.x), where the go version should now be used.

Golangci-lint is also updated with some Go warning.

The protobuf generation uses a new docker image with golang 1.23.


jefft0 (approved):
LGTM. The failed "Android / Build for Android" is expected an will be fixed in the future.

PR #5114, opened 2026-05-29
Fix/android ios ci

D4ryl00:
## Summary

- **Fix Android CI**: `eas build --output` was pointing to the workspace directory instead of a file path, causing the build to fail when trying to overwrite the checkout directory with the `.aab` file
- **Fix iOS CI**: EAS local build failed during cleanup because gomobile's internal Go module cache creates read-only files (`0444`); fixed by adding `chmod -R u+w` to the iOS gomobile Makefile targets so files are writable before EAS cleans up its temp directory
- **Wire semantic-release to mobile builds**: replace the old docker-based semantic-release action with `cycjimmy/semantic-release-action@v4` and integrate `semantic-release-expo` so each release automatically bumps `version`, `ios.buildNumber`, and `android.versionCode` in `app.json`; android/ios production builds are then dispatched via `gh workflow run` only when a new release is published and relevant files changed
- **Remove debug logs** in `notification.context.tsx` and `messenger.hooks.ts`

## Details

**Version management** (`eas.json`, `app.config.ts`, `app.json`, `.releaserc`):
- `semantic-release-expo` now owns versioning — it writes `version`, `buildNumber`, and `versionCode` to `app.json` on each release, committed back by `@semantic-release/git`
- `app.config.ts` no longer imports version from `package.json`; it forwards `buildNumber` and `versionCode` from the Expo config context (which merges `app.json`)
- `APP_ENV` env var added to `eas.json` profiles so `app.config.ts` can select the right bundle identifier per environment

**CI flow** (`android.yml`, `ios.yml`, `release.yml`):
- Android and iOS workflows no longer trigger on master push; they now only run via `workflow_dispatch` (dispatched by `release.yml` after a successful semantic release)
- Path filtering is preserved: `dorny/paths-filter` in `release.yml` checks that Go or `berty-bridge-expo` files actually changed before dispatching each platform's build
- `--profile production` made explicit on all `eas build` / `eas upload` commands

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -71,7 +71,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ hashFiles('go/**/go.sum') }}
+          key: ${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ hashFiles('go.sum') }}
           restore-keys: ${{ runner.os }}-go-${{ env.GO_VERSION }}-
 
       - name: Cache expo module node modules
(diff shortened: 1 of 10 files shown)


jefft0 (approved):
Looks good. Need to merge it to test.

PR #5134, opened 2026-06-17
Port BLE fixes and improvements

leskunberg:
# Port BLE fixes and improvements

## Motivation

We run a Berty fork at CoHunt. The bluetooth data exchange is extremely important to our users in circumstances with no cell service. This PR is meant to enhance and bridge communication between iOS and Android users.  Feedback is welcome!

## Summary

Adds a **native iOS CoreBluetooth BLE proximity transport** for Berty (iOS previously had no native BLE driver), ports the same reliability/safety model into the existing **Android BLE driver**, and **improves the iOS connectivity driver**.

~4,600 lines across iOS (`js/ios`), expo (`berty-bridge-expo/ios`), and Android (`js/android`). No Go changes — the iOS transport binds to the existing gomobile `Bertybridge` proximity protocols and is wire-compatible with the Android BLE driver.

### iOS — native BLE proximity transport (new)
- **`BertyBLEDriver.swift`** — CoreBluetooth engine acting as both GATT central and peripheral; custom PID handshake, chunked data over a WRITER characteristic, per-peer connection/handshake timeouts, lazy Bluetooth init, CoreBluetooth state restoration. Wire-compatible with Android (service `0x4240`, PID char `0x4241`, WRITER char `0x4242`, `EOD` marker, `[PSM][PID]` framing; iOS pins PSM=0 as it has no L2CAP).
- **`BertyDriverBridges.swift`** — Go wiring: `BertyProximityDriverBridge` (`BertybridgeProximityDriverProtocol`, `protocolName "ble"`, `protocolCode 0x0042`, `defaultAddr "/ble/Qmeeee…"`); pulls the transport via `BertybridgeGetProximityTransport("ble")` and forwards driver callbacks into the Go transport. Also an MDNS-locker bridge and a background-task shim.
- **`BertyProximityManager.swift`** — higher-level BLE + Bonjour/Wi-Fi coordinator (see Known limitations).
- **Supporting infra:** `BertyConfiguration` (constants incl. the interop-critical BLE UUIDs), `BertyQueueManager` (named queues + thread-safe collections), `BertyErrors` (error taxonomy + recovery), `BertyNetworkDriver` (interface/reachability introspection), `ActivityBackgroundState`, and `BertyLogger` updates (typo fix, `Comparable` levels, production log gating).

### iOS — connectivity driver improvements
- Reworked `ConnectivityDriver` to share `BertyBLEDriver`'s `CBCentralManager` when available (else owns one — strong ref, power-alert suppressed), with a thread-safe handler list, background-safe Go callbacks, fresh Bluetooth-state reads, and an iOS<12 reachability fallback. Applied to both `js/ios` and `berty-bridge-expo` (expo owns its own manager, since it has no BLE driver).

### Android — BLE reliability hardening
- Real timeouts replacing unbounded/mis-unit waits (notably a `GattServer` write that effectively waited ~2.78h due to a SECONDS/MS unit bug; several unbounded `await()` calls now bounded and tear down stale connections).
- Concurrency hardening (`volatile` + dedicated locks in `BleInterface`, `AtomicLong` in `BleQueue`, `ConcurrentHashMap` in `Scanner`), stoppable L2CAP accept/read threads with bounded `join()` + socket-leak fixes, null-safety guards, and a 64KB inbound-buffer cap (OOM protection).

## Wire compatibility
BLE service/characteristic UUIDs, the `EOD` marker, and `[PSM][PID]` framing match the Android driver so iOS↔Android proximity works. `protocolCode 0x0042` / `protocolName "ble"` / `defaultAddr` must stay in sync with the Go and Android sides.

## Known limitations / open questions for reviewers
- **Logging is gated to ≥WARN unless `initializeDebugMode(true)` is called at startup**, which isn't wired in yet (affects both the new iOS and Android loggers) — INFO/DEBUG is currently suppressed in all builds. Needs a startup call.
- **`BertyBLEDriver` reads some mutable dictionaries off its serial `bleQueue` without locking** in a few public methods — potential data races to tighten.
- **`BertyProximityManager` and `BertyNetworkDriver` appear auxiliary** — not on the live Go transport path; the proximity manager's multicast connect/send paths and the net driver's `updateAddr` are stubs. Confirm whether to wire them in or trim.
- BLE initiator tie-break uses a 4-char PID-suffix comparison (collision-prone at scale).
- Android: operations now return `false`/`disconnect()` on timeout (a slow-but-alive peer >30s gets torn down); the 64KB single-accumulation cap silently drops larger windows.

### DIFF
--- a/berty-bridge-expo/ios/Sources/Bridge/ConnectivityDriver.swift
+++ b/berty-bridge-expo/ios/Sources/Bridge/ConnectivityDriver.swift
@@ -5,135 +5,281 @@
 //  Created by u on 01/02/2023.
 //
 
-import SystemConfiguration
-import CoreBluetooth
-import CoreTelephony
 import Foundation
-import Bertybridge
+import UIKit
 import Network
+import SystemConfiguration
+import CoreTelephony
+import CoreBluetooth
+import Bertybridge
 
-class ConnectivityDriver: NSObject, BertybridgeIConnectivityDriverProtocol, CBCentralManagerDelegate {
-    let logger: BertyLogger = BertyLogger("tech.berty.ConnectivityDriver")
-    let queue = DispatchQueue.global(qos: .background)
-    let pathMonitor = NWPathMonitor()
-    var centralManager: CBCentralManager!
+// Bridges iOS network and Bluetooth state to Go's netmanager.
+//
+// Network reachability comes from NWPathMonitor, with an SCNetworkReachability
+// fallback for iOS < 12. Bluetooth state is observed through a CBCentralManager we
+// own (the expo bridge has no shared BLE driver to borrow one from); we keep a
+// strong reference so it is not deallocated and suppress the system power alert
+// since we only observe state.
+//
+// The connectivity constants (state/net/cellular) are exported by the Bertybridge
+// framework from Go's netmanager/connectivity.go — never redefine them locally.
+class ConnectivityDriver: NSObject, BertybridgeIConnectivityDriverProtocol {
+    let logger = BertyLogger("tech.berty.connectivity")
 
-    var state: BertybridgeConnectivityInfo
-    var handlers: [BertybridgeIConnectivityHandlerProtocol] = []
+    private var pathMonitor: NWPathMonitor?
+    private let monitorQueue = DispatchQueue(label: "tech.berty.connectivity.monitor")
+
+    // Owned manager: kept as a strong reference so it is not deallocated.
+    private var centralManager: CBCentralManager?
+
+    private let networkInfo = CTTelephonyNetworkInfo()
+
+    private var isMonitoring = false
+    private var currentPath: NWPath?
+    private var bluetoothState: CBManagerState = .unknown
+
+    // Thread-safe handler storage. Go callbacks run on a dedicated serial queue
+    // (not DispatchQueue.global(), which causes stack-alignment issues in the Go
+    // runtime) when the app is backgrounded.
+    private let handlerQueue = DispatchQueue(label: "tech.berty.connectivity.handlers", attributes: .concurrent)
+    private let goCallbackQueue = DispatchQueue(label: "tech.berty.connectivity.go-callbacks")
+    private var handlers: [BertybridgeIConnectivityHandlerProtocol] = []
 
     override init() {
-        self.logger.debug("Init")
-      
-        self.state = BertybridgeConnectivityInfo()!
-
         super.init()
-      
-        self.centralManager = CBCentralManager(delegate: self, queue: nil)
-
-        self.pathMonitor.pathUpdateHandler = { [weak self] path in
-            self!.logger.debug("Network state changed")
-          
-            self!.updateNetworkState(path)
-
-            for handler in self!.handlers {
-                handler.handleConnectivityUpdate(self!.state)
-            }
-        }
-        self.pathMonitor.start(queue: self.queue)
+        self.startMonitoring()
     }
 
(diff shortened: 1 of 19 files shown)


D4ryl00:
Hello, and thank you for your contribution!

I have a few comments and questions before we can consider it.

The `js/` folder is deprecated. The Android BLE driver is now provided to the app through an Expo native module — see `./berty-bridge-expo/android/src/main/java/tech/berty/bertybridgeexpo/gobridge/bledriver`.

For iOS, a BLE driver already exists, but at the weshnet level: https://github.com/berty/weshnet/tree/main/pkg/ble-driver — it's an Objective-C driver. We deliberately don't want a Swift version for the Berty Messenger, because we want to provide the iOS BLE driver directly to weshnet as a transport. The only way to do that is to compile it with cgo, which requires an Objective-C driver.

You mention in the PR that the new driver is wire-compatible with the Android BLE driver. My main question is therefore: do the reliability/hardening changes you made to the Android driver keep it compatible with the Objective-C driver in weshnet?

Given the above, would you be open to porting your iOS improvements to the Objective-C driver in weshnet, and rebasing the Android changes onto the Expo native module rather than `js/android`?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/berty/weshnet/tree/main/pkg/ble-driver">weshnet/pkg/ble-driver at main · berty/weshnet</a></strong></div><div>Async Mesh Network Protocol for Extreme Communication -- Innovative, Resilient, and Decentralized - berty/weshnet</div></blockquote>

leskunberg:
> Hello, and thank you for your contribution!
> 
> I have a few comments and questions before we can consider it.
> 
> The `js/` folder is deprecated. The Android BLE driver is now provided to the app through an Expo native module — see `./berty-bridge-expo/android/src/main/java/tech/berty/bertybridgeexpo/gobridge/bledriver`.
> 
> For iOS, a BLE driver already exists, but at the weshnet level: https://github.com/berty/weshnet/tree/main/pkg/ble-driver — it's an Objective-C driver. We deliberately don't want a Swift version for the Berty Messenger, because we want to provide the iOS BLE driver directly to weshnet as a transport. The only way to do that is to compile it with cgo, which requires an Objective-C driver.

That makes sense. The separation of concerns seems to point to weshnet as the transport as you say.

> 
> You mention in the PR that the new driver is wire-compatible with the Android BLE driver. My main question is therefore: do the reliability/hardening changes you made to the Android driver keep it compatible with the Objective-C driver in weshnet?

Structurally yes, they are wire-compatible - but I believe this needs a live smoke test since a static inspection really only goes so far. I'm curious on the ObjC drivers max message size. If the EOD marker is outside the 64KB accumulation OOM guard, Android would drop the message. 


> 
> Given the above, would you be open to porting your iOS improvements to the Objective-C driver in weshnet, and rebasing the Android changes onto the Expo native module rather than `js/android`?

Absolutely! That seems like a reasonable next step.

> 
> >  GitHub**[weshnet/pkg/ble-driver at main · berty/weshnet](https://github.com/berty/weshnet/tree/main/pkg/ble-driver)**Async Mesh Network Protocol for Extreme Communication -- Innovative, Resilient, and Decentralized - berty/weshnet



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/berty/weshnet/tree/main/pkg/ble-driver">weshnet/pkg/ble-driver at main · berty/weshnet</a></strong></div><div>Async Mesh Network Protocol for Extreme Communication -- Innovative, Resilient, and Decentralized - berty/weshnet</div></blockquote>

PR #5138, opened 2026-06-19
fix(mobile): stabilize Expo mobile build (chat, notifications, Android edge-to-edge, iOS lifecycle, logging)

D4ryl00:
## Overview

Stabilizes the Expo (SDK 53/54, new architecture) mobile build: fixes a set of regressions that surfaced after the migration (message ordering, paging, notifications, deep links, Android edge-to-edge layout, an iOS watchdog kill) plus two backend message-delivery bugs and logging/diagnostics improvements.

The branch is organized as small, self-contained commits — reviewing commit-by-commit is the easiest path. Below is what each area fixes and how to verify it.

---

## Chat & messaging

**`fix(chat): correct message order and live updates for FlashList v2`**
FlashList v2 dropped the `inverted` prop and ignores `estimatedItemSize`, so messages rendered newest-first at the top and didn't refresh until touched. Inversion is emulated with a `scaleY(-1)` wrapper + per-cell counter-flip, and `maintainVisibleContentPosition.autoscrollToTopThreshold` keeps new messages flowing in.
*Verify:* open a conversation → messages are oldest→newest bottom-anchored; receive a message → it appears at the bottom without touching the list.

**`fix(chat): clear loading spinner after paging completes`**
The footer spinner (top of the inverted list) stuck forever after the first page because the cleanup effect's guard was false exactly when it mattered. Now cleared in a `finally`.
*Verify:* scroll up to page older messages → spinner disappears once the page loads.

**`fix(chat): only treat ErrNotFound as end-of-history when paging`**
`fetchMore` caught *every* error as "reached the beginning", so a transport failure silently stopped paging. Now only `ErrNotFound` marks end-of-history; other errors are logged and paging stays enabled.

**`fix(messenger): parse interactions with numeric AppMessage type`**
`parseInteraction` assumed the protobuf JSON enum string form; optimistic local interactions (built in `ChatFooter`) pass the numeric enum, so `substring` threw and the message was stored as `Undefined` and never rendered. Numeric `type` is now normalized to the enum-name string. Unit tests cover string/numeric/missing cases.
*Verify:* send a message → it renders immediately (optimistic) and persists.

**`fix(messenger): keep group message delivery alive across stream drops`** *(Go backend)*
The group message subscription goroutine returned on the first stream error, permanently stopping interaction indexing for that group with no reconnect and no backfill — messages went missing even across restarts (a normal restart re-subscribes with `SinceNow` and only replays the log on schema migration). This commit makes delivery survive drops:
- Reconnect on `Recv` errors with immediate-then-exponential backoff, resuming from the last received message id (`SinceId`) so messages from the outage are backfilled. The cursor is seeded from the last stored interaction so a drop before the first live message can still resume.
- On (re)subscription, a one-shot bounded catch-up (`GroupMessageList` with `SinceId` = last indexed, `UntilNow`) replays any pre-existing log gap up to the head and stops; it does not tail. `AddInteraction` is idempotent (Cid PK), so overlap with the live stream is a harmless no-op.
- Decode failures skip the message instead of tearing down the stream.

Unit tests cover reconnect/backfill, seeded-cursor, cancellation, the bounded drain, and the reconciliation request shape (`SinceId == cursor`, `UntilNow`, not `SinceNow`) via injectable stream/connector seams.
*Why this is hard to verify by hand:* it only triggers when a stream drops or a message was lost in a prior session; the tests assert the reconnect/replay behavior and request shape.

---

## Notifications & deep links

**`fix(notifications): suppress in active chat, route taps, dismiss on open`**
- No notification (push or in-app) for the conversation currently on screen.
- Response handler hardened (missing `break`, default guard, try/catch) so a malformed notification no longer crashes into the restart screen.
- Tap routes to the right conversation via `convPK` regardless of path, including cold-start launches (`getLastNotificationResponseAsync`) once the messenger client is ready.
- Conversation tray notifications are cleared when the chat is opened.
- Android: the FCM tap intent now explicitly launches the main activity (was an implicit `ACTION_VIEW` that did nothing when the deep link was empty).
*Verify:* background the app, receive a message, tap the notification → lands in the correct conversation (cold start too); with the chat open, no notification is posted.

**`fix(deeplink): handle launch URL once and navigate from an effect`**
Expo dev-client launch URLs are ignored (not Berty links). The persisted launch URL is processed once per session (module guard) — previously every `DeepLinkBridge` instance re-read it, causing a navigate→parse loop. Message-link navigation moved out of render into an effect to stop the "update a component while rendering" churn that froze the conversation screen.

---

## Android edge-to-edge & layout

Under the new-arch edge-to-edge mode, `react-native` `StatusBar` `backgroundColor`/`translucent` are deprecated and apply a contrast scrim, formSheets behave differently, and the keyboard overlaps content.

- **`fix(android): edge-to-edge status bar via expo-status-bar`** — switch every screen to `expo-status-bar` (`style` prop only), which routes through the edge-to-edge SystemBars, so none of them trigger the Android scrim.
- **`fix(android): white (iOS-like) navigation bar`** — `enforceContrast=false` + dark icons so the white content shows through the transparent bar.
- **`fix(keyboard): avoid keyboard with react-native-keyboard-controller`** — replaces the broken `react-native-android-keyboard-adjust` import (not installed, broke the bundle) and the iOS-only `KeyboardAvoidingView` with a deterministic reanimated bottom-padding driven by the keyboard's animated height. *Verify:* open a chat on Android, focus the input → input bar rides on top of the keyboard.
- **`fix(settings): keep formSheet content clear of the Android status bar`** — `useTopInset` (Android-only top inset) on settings screens that render as headerless formSheets.
- **`fix(devtools): enable nested scrolling`** — `nestedScrollEnabled` so the DevTools page scrolls past the dropdowns on Android.
- **`fix(navigation): present SelectNode full-screen instead of an empty/jumping formSheet`** — as the initial route a root formSheet renders empty on Android and size-jumps on iOS; presented as a full-screen card on both, with the top safe-area inset applied manually.
- **`fix(onboarding): present Permissions full-screen to stop layout jump`** — formSheet sized to JS-computed content height and reflowed while the JS thread was busy during account open; `fullScreenModal` uses the fixed native height.
- **`fix(ui): seed SafeAreaProvider with initial window metrics`** — `initialWindowMetrics` so insets are correct on the first frame instead of starting at 0 and jumping.
- **`fix(permissions): keep dismiss button clear of the home indicator`** — add the bottom safe-area inset to the permission card.
- **`fix(scan): keep QR scanner open after granting camera permission`** — the permission screen ran deny()+goBack() twice, popping the scanner; now it resolves once.

---

## iOS

**`fix(ios): run blocking lifecycle bridge calls off the main thread`**
`HandleState`/`willTerminate` block until Go background tasks finish (e.g. unsubscribing from every group), which can exceed iOS's ~5s main-thread transition budget and trip the watchdog (`0x8BADF00D` SIGKILL). They now run on a serial off-main queue under a background-task assertion (iOS grants ~30s).
*Why hard to verify:* it's a race against the watchdog timeout; reproduces as an occasional SIGKILL on backgrounding/terminating with many active groups.

---

## Logging & diagnostics

**`fix(logs): correct default log filter and enable info+ in production`**
The default filter used the `bertybridge.bty*` namespace, which never matches the Go core loggers (`bty.*`), so Go debug/info logs were silently dropped; the trailing `error+:*` also parsed as a dead namespace pattern. Now uses `bertybridge*,bty*` with `-*.grpc`, and picks verbosity by build variant via `expo-application` (debug+ for debug/staff/yolo, info+ for production — which previously logged nothing). Keyed on `applicationId` so a bundled production build is distinguished from a staff release.

**`fix(logs): apply configured filter to on-device file logs`**
On-device Go file logs defaulted to debug+ regardless of build; now follow the configured verbosity via `--log.file-filters`.

**`feat(tool): add mobile-logs script`**
`tool/mobile-logs/get-mobile-logs.sh` collects the JS/RN console, Go core logs and on-device JSON log files into a zipped bundle for bug reports (Android via `adb`, iOS via `xcrun simctl`/`devicectl`). File logs only from debuggable builds.

**`fix(settings): pass gomobile git version to app`** — surface the gomobile build version in the app.

---

## Misc

- **`fix(bridge): unify on a single GoBridge, drop the stale null wrapper`** — two objects were both named `GoBridge`; the local `@berty/native-modules/GoBridge` wrapper resolved `NativeModules.GoBridge` which is `null`, so `closeBridge` (account close/switch/logout/delete) threw and left account deletion stuck on "Deleting account…". All importers now point at the `berty-bridge-expo` package; the wrapper is deleted. *Verify:* delete an account → it completes instead of hanging.
- **`fix(settings): prevent system info clipping`** / **`show Long fields as numbers`** / **`fix(progress): render Long progress counters as text`** — protobufjs now resolves the `long` library after the Expo bump, so int64 fields decode as `Long` objects; they're stringified for display (previously rendered as `{low, high, unsigned}` or threw "Objects are not valid as a React child").

---

## Docs

**`docs: deprecate ./js in favor of berty-bridge-expo`**
The `./js` React Native app is superseded by `berty-bridge-expo` (Expo) and will be removed soon. Adds a deprecation banner to `js/README.md` and points the repo README at `berty-bridge-expo` as the current home of the mobile app.

---

## CI & tests

**`test(replication): de-flake TestReplicationService_Flow`** *(Go backend)*
The flake is ultimately a **go-libp2p-pubsub connect/disconnect race**: on a fast reconnect the new-peer event is deduped against the not-yet-removed stale peer entry, so the reconnecting peer's subscriptions are never re-propagated and it ends up in 0 topics despite a healthy `Connected` connection — an upstream race, amplified by mocknet, that can't be fully removed at the test level. The test now uses the event-based `pubsubraw` like production, a bounded reconnect retry loop for the offline-peer catch-up, and is marked `FilterStability(Flappy)` so it runs in `go.flappy-tests` (`testman -retry=10`) instead of the blocking stable lane. Root cause + full investigation are written up for an upstream issue.

**`test(replication): fix orphaned TestReplicationService_GroupRegister`** *(Go backend)*
Marked `FilterStability(Flappy)` but no flappy lane covered `bertyreplication`, so it never ran — and had rotted into a nil-deref panic: the reopen step reused a context the test had already cancelled, failing the group resubscription. Now uses a fresh context. Both bertyreplication flappy tests are renamed to the `TestFlappy*` prefix so the lane picks them up via `-run ^TestFlappy`, like `bertymessenger`/`bertybot`.

**`fix(ci): wait for pprof server before scraping in benchmark job`**
The benchmark job slept a fixed 2s before curling the pprof server, which sometimes wasn't up yet, failing the job. Replaced with `curl --retry-connrefused`.

**`ci: remove deprecated ./js folder workflows`**
Removes the CI tied to the deprecated `./js` app (superseded by `berty-bridge-expo`) so it stops failing on unrelated PRs; `release.yml` now publishes `@berty/api` from `berty-bridge-expo/mobile/src/api`.

---

## Notes

- `SelectNode.test.tsx`'s snapshot covers the presentation change and would need `jest -u`, but the screen snapshot suite is currently non-runnable in this repo (unrelated test-harness issues).


jefft0 (approved):
Tested on Android and iOS simulators:
* fix(chat): correct message order and live updates for FlashList v2
* fix(chat): clear loading spinner after paging completes

Confirmed getting notifications on physical iPhone.

Tested on Android simulator:
* fix(android): white (iOS-like) navigation bar
* fix(keyboard): avoid keyboard with react-native-keyboard-controller
* fix(settings): keep formSheet content clear of the Android status bar
* fix(devtools): enable nested scrolling
* fix(navigation): present SelectNode full-screen instead of an empty/jumping formSheet
* fix(onboarding): present Permissions full-screen to stop layout jump
* (others are internal)

PR #5144, opened 2026-06-29
fix(ci): resolve bertymessenger test data race and Android release disk exhaustion

D4ryl00:
## Summary

Fixes two unrelated CI failures observed on `master`.

### 1. `test(bertymessenger)`: data race in account update tests

The `Stable tests (macOS)` job failed under the `-race` flag (passes without it). `TestAccountUpdate` and `TestFlappyAccountUpdateGroup` read the `TestingAccount.account`/`members` fields directly while the background goroutine spawned by `ProcessWholeStream` writes them under `processMutex`, causing a detected data race.

Fix: route those reads through the existing mutex-protected `GetAccount()` / `GetMember()` accessors.

Verified locally with the race detector:
- `TestAccountUpdate` — 5/5 PASS, no data race
- `TestFlappyAccountUpdateGroup` (`TEST_STABILITY=flappy`) — 3/3 PASS, no data race

### 2. `ci(android)`: free runner disk space before the release AAB build

The release (master) Android build runs a full multi-ABI `eas build --local` that compiles all native C++ modules (`react-native-worklets`, `expo-modules-core`) for x86, x86_64, armeabi-v7a and arm64-v8a. This exhausts the ~14 GB `ubuntu-latest` disk, failing with `No space left on device` during the clang link step (surfacing as `Bus error (core dumped)`).

PR builds only run a single-ABI compile check, so they never hit this — which is why it failed on `master` but not in PRs.

Fix: free unused preinstalled toolchains (.NET, GHC, Boost, Swift) and prune Docker images before the build, gated to the real-build path. The Android SDK/NDK are kept intact.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -27,6 +27,18 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
+      - name: Free disk space
+        if: inputs.app-json-artifact != ''
+        run: |
+          echo "Disk space before cleanup:"
+          df -h /
+          # Remove large preinstalled toolchains unused by the Android build.
+          # Keep /usr/local/lib/android (SDK/NDK) — the build needs it.
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/.ghcup /usr/local/share/boost /usr/share/swift
+          sudo docker image prune --all --force || true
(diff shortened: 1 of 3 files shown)


jefft0 (approved):
Looks good. Need to merge to test the release workflow.

PR #5151, opened 2026-07-02
fix(bridge-expo): derive iOS App Group per build variant

D4ryl00:
## Problem

The iOS App Group was hardcoded to `group.tech.berty` for **every** build variant.

The Go bridge keeps two roots on iOS: `appRootDir` = the app-sandbox `Documents/berty` (per-install), and `sharedRootDir` = the iOS **App Group** container (shared by group id). `ListAccounts`/account selection read `sharedRootDir`, but the open-time existence check (`accountExists`) reads `appRootDir`.

When two variants are installed side-by-side (e.g. production `tech.berty.ios` and debug `tech.berty.ios.debug`), they shared the App Group container. The debug build then saw production's account list via `ListAccounts` but not its data via the existence check, failing with:

```
ErrBertyAccountOpenAccount(#5006): ErrBertyAccountDataNotFound(#5007)
```

and hanging on the opening loader.

## Fix

Derive the App Group from the bundle-identifier variant instead of hardcoding it:

| Bundle identifier | App Group |
|---|---|
| `tech.berty.ios` (production) | `group.tech.berty` |
| `tech.berty.ios.debug` (development) | `group.tech.berty.debug` |
| `tech.berty.ios.staff` (preview) | `group.tech.berty.staff` |

- New `plugin/src/appGroup.ts` — `getAppGroupID(bundleIdentifier)` helper.
- Wired into `withIosEntitlements.ts` (app-groups entitlement) and `withIosPlist.ts` (`appGroupID` Info.plist key — also used as the keychain access group for the keystore).
- `withIosPush.ts` rewrites the copied NotificationService `ProdNS.entitlements` + `NotificationService-Info.plist` for non-prod variants so the extension shares the **same** per-variant container (needed for push decryption and the keychain-backed keystore).
- `mobile/app.config.ts` derives the NSE App Group the same way for EAS builds (small inline copy of the helper — Expo evaluates `app.config.ts` in isolation via sucrase and can't reliably import the plugin's `.ts` source or its gitignored `build/` output).

This mirrors the original app's historical scheme (`group.tech.berty.dev` / `group.tech.berty.entreprise`).

Android is unaffected (per-package `getNoBackupFilesDir`, no App Groups; variant packages already differ).

## Tested

Tested on iOS with two different app variants. They didn't share their DB anymore.

### DIFF
--- a/berty-bridge-expo/mobile/app.config.ts
+++ b/berty-bridge-expo/mobile/app.config.ts
@@ -77,7 +77,7 @@ export default ({ config }: ConfigContext): ExpoConfig => {
 									bundleIdentifier: `${bundleIdentifier}.NotificationService`,
 									entitlements: {
 										"com.apple.security.application-groups": [
-											"group.tech.berty",
+											getAppGroupID(bundleIdentifier),
 										],
 										"com.apple.developer.associated-domains": [
 											"applinks:berty.tech",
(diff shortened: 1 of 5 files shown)


jefft0 (approved):
My iOS simulator had a previously installed Berty Debug. I built Berty Debug in this branch and it started on the iOS simulator with:
```
unable to init bridge:  [Error: ErrDBOpen(#124): failed to ping database: SQL logic error]
[React] 'unable to init bridge: ', { [Error: ErrDBOpen(#124): failed to ping database: SQL logic error] code: 'ERR_UNEXPECTED' }
```
This is to be expected with the database name change. I deleted Berty Debug from the iOS simulator and reinstalled. It works. After merge, we will confirm that installing the update from the App Store keeps using the same database on the iPhone.

PR #5207, opened 2026-08-12
fix: restore the Stargazers over time chart in the README

OctoBored:
The "Stargazers over time" chart in the README is currently broken and does not render due to the GitHub stargazer API restrictions. This change updates the link so the chart works again and stays up to date with the project's growing community.


jefft0:
@OctoBored , it works for me. Can you check again? What browser are you using?
https://github.com/berty/berty/blob/master/README.md#stargazers-over-time

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/berty/berty/blob/master/README.md">berty/README.md at master · berty/berty</a></strong></div><div>Berty is a secure peer-to-peer messaging app that works with or without internet access, cellular data or trust in the network - berty/berty</div></blockquote>

